PR #7, opened 2025-05-16
Add example using mem0_memory tool

deshraj:
## Description

Adding an example of using `mem0_memory` as a tool with strands Agent.

## Type of Change
- New content addition


deshraj:
Closing this since we are proceeding with https://github.com/strands-agents/docs/pull/11

PR #16, opened 2025-05-19
Add examples for therapist agent and personalized tutor agent

deshraj:
## Description

Added two examples:
- Therapist Agent
- Personalized Tutor Agent


deshraj:
> IMO it belongs more in the samples repo, since they are usecases. Here in docs, perhaps we should juse showcase simple usage (how-to).

Yeah makes sense. Wasn't aware if there is a `samples` repo.

PR #43, opened 2025-05-22
Add lychee link checker as GitHub wrokflow

didier-durand:
## Description

Adding a GitHub workflow running Lychee action to have a daily check of links (to find 404. 403, etc.)  in the documentation

## Type of Change

- Other (please describe):

See above: adding a workflow to check validity of url links

## Motivation and Context

As you'll see on first run (or in my fork), there are already some incorrect links in the doc. This workflow will detect them so that they can get fixed.

## Areas Affected

Global quality of documents

## Checklist

- [ ] I have tested the documentation locally using `mkdocs serve`
- [ ] Images/diagrams are properly sized and formatted

## Additional Notes

The 2nd step of the workflow automatically opens an issue in the project when incorrect links have been found. To get it to work in the repo, proper token will have to be granted to this workflow. 

See https://github.com/lycheeverse/lychee-action : *"When used in conjunction with [Create Issue From File](https://github.com/peter-evans/create-issue-from-file), issues will be opened when the action finds link problems (make sure to specify the issues: write permission in the [workflow](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions) or the [job](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions))"*

See also https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -5,6 +5,7 @@ on:
   push:
     paths:
       - '.github/workflows/check-links.yaml'
+      - '.lycheeignore'
 
   repository_dispatch:
 
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -1,0 +1,29 @@
+name: Links
+
+on:
+
+  repository_dispatch:
+
+  workflow_dispatch:
+
+  schedule:
+    - cron: "00 09 * * *"
+
+jobs:
+  check_links:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+
+      - name: Create Issue From File
+        if: env.lychee_exit_code != 0
+        uses: peter-evans/create-issue-from-file@v5
+        with:
+          title: Link Checker Report
+          content-filepath: ./lychee/out.md
+          labels: report, automated issue
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -2,6 +2,10 @@ name: Links
 
 on:
 
+  push:
+    paths:
+      - '.github/workflows/check-links.yaml'
+
   repository_dispatch:
 
   workflow_dispatch:
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,0 +1,4 @@
+https://locahost*
+https://127.0.0.1*
+http://locahost*
+http://127.0.0.1*


.github/workflows/check-links.yaml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+
+      - name: Create Issue From File
```
zastrowm, line 28: Let's skip creating issues, we can just use the checker on PRs

.github/workflows/check-links.yaml
```diff
@@ -0,0 +1,34 @@
...
+    paths:
+      - '.github/workflows/check-links.yaml'
+      - '.lycheeignore'
+
+  repository_dispatch:
+
+  workflow_dispatch:
+
+  schedule:
+    - cron: "00 09 * * *"
```
zastrowm, line 15: Unless there's a reason not to, let's switch to running this on every PR.

It looks like there's a cache to mitigate rate limiting problems: https://github.com/lycheeverse/lychee-action?tab=readme-ov-file#utilising-the-cache-feature if we need it

.lycheeignore
```diff
@@ -0,0 +1,4 @@
+https://locahost*
+https://127.0.0.1*
+http://locahost*
+http://127.0.0.1*
```
zastrowm, line 4: From the report generated (https://github.com/didier-durand/strands-agents-docs/actions/runs/15179282564):

 - It looks like some domains are blocking the checker - that's fine, but let's ignore those (openai, www.oecd.org)
 - http://localhost:16686/ should be ignored as well; not sure why that's not being caught by the list above

.github/workflows/check-links.yaml
didier-durand: Ok. I removed this step for tickets

.github/workflows/check-links.yaml
didier-durand: Added running on every PR. But kept the cron to have at least a daily check in case remote URLs disappear. Also, created a cache with max age 1 d so that every external url is verified max once per day.

.lycheeignore
didier-durand: Ok, I used syntax of https://github.com/opensafely/documentation/blob/main/.lycheeignore as examples to fix current version.

The 2 remaining errors on this repo seem valid: I get 404 with my browser for them

didier-durand:
I guess that this is now good to merge.

Let me know if you want me to open a new PR for the 2 incorrect links in the repo or if you prefer to do it on you own.

.github/workflows/check-links.yaml
zastrowm: Sorry I should have been more clear - I was requesting that the workflow be switched to only run on pull requests. But you do make a good point about the cron job.

If we keep the cron job, please:

 - Remove push triggers (we only need PRs + cron)
 - Remove the `repository_dispatch` permission
 - Update the permission of the workflow to be `contents: read` only at the workflow level: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#defining-access-for-the-github_token-permissions
 - Update the pull_request trigger types to match the SDK: https://github.com/strands-agents/sdk-python/blob/main/.github/workflows/test-lint-pr.yml#L4

zastrowm (changes requested):
> Let me know if you want me to open a new PR for the 2 incorrect links in the repo or if you prefer to do it on you own.

A PR would be greatly appreciated, but we can pick it up too! You can include it in this PR if you want as well

didier-durand:
Hi,

 I made what should be final commit to match the requirements of https://github.com/strands-agents/docs/pull/43#discussion_r2103478237

1. removed push triggers
2. removed repository dispatch trigger
3. removed manual dispatch trigger
4. added permission token for read only
5. aligned the pull_request trigger to match example given above

It should be good to merge now!

Thanks for the patience

Didier

(I'll come back with PR for error links that remain  - like 404 -  when the job is executed)

zastrowm (approved):

didier-durand:
@zastrowm : thank you for adding the last commit.

ryanycoleman:
Thank you @didier-durand!

PR #91, opened 2025-06-16
feat: Add workflow to build docs on every PR

zastrowm:
## Description

To ensure that things are actually working

## Type of Change

New content addition

## Motivation and Context

We're not actually building the docs as part of each PR, which means we could break things

## Checklist

- [X] Images/diagrams are properly sized and formatted

## Additional Notes


.github/workflows/build-test.yaml
```diff
@@ -12,12 +11,25 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          pip install -r requirements.txt
+      - name: Build docs
+        run: |
+          mkdocs build
```
Unshure, line 25: Should we build strict here?
```
mkdocs build --strict
```

Unshure (approved):

.github/workflows/build-test.yaml
zastrowm: I tried that - we have build failures right now with `--string` so I didn't include it.

That'd be a good follow-up though

PR #137, opened 2025-07-13
[BUG] fix docs build due to misaligned strands versions in docs

yonib05:
## Description
Docs build is breaking due to the API ref not finding the right class due to misaligned Strands versions. I am defaulting it to the latest PyPI version to avoid this issue in the future.

## Type of Change
- Bug fix

## Motivation and Context
Defaulting to the latest PyPI so docs will always build API ref to latest

## Checklist

- [x] Images/diagrams are properly sized and formatted
- [x] All new and existing tests pass

## Additional Notes

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of your choice.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,4 +4,4 @@ mkdocs-macros-plugin~=1.3.7
 mkdocs-material~=9.6.12
 mkdocstrings-python~=1.16.10
 mkdocs-llmstxt~=0.2.0
-strands-agents~=0.2.0
+strands-agents


ryanycoleman:
nice find

awsarron (approved):

requirements.txt
```diff
@@ -4,4 +4,4 @@ mkdocs-macros-plugin~=1.3.7
 mkdocs-material~=9.6.12
 mkdocstrings-python~=1.16.10
 mkdocs-llmstxt~=0.2.0
-strands-agents~=0.2.0
+strands-agents
```
zastrowm, line 7: Not that we have to be careful with branch builds now as otherwise they'll always pull the latest. But maybe that's just needs to be part of our branch workflow

We also have a reference to 0.2.0 for publishing in the workflow FWIW

yonib05: Yeah, on the new site process I built locally, I dynamically get those as well.

I am happy to just bump this to 0.3.0 to fix, but I figured we rarely publish old versions of docs.

zastrowm: I concur - If we could make both dynamic that'd be ideal and that's what we should work towards instead of hardcoding

zastrowm (approved):

PR #228, opened 2025-08-28
tool executors

pgrayy:
## Description
Adding documentation for `ToolExecutor` usage. This is a follow up to https://github.com/strands-agents/sdk-python/pull/658

## Type of Change
- New content addition

## Motivation and Context
Provides usage details for `ConcurrentToolExecutor` and `SequentialToolExecutor`.

## Areas Affected
- `user-guide/concepts/tools/executors.md`
- `api-reference/tools.md`

## Checklist
- [x] Images/diagrams are properly sized and formatted
- [x] All new and existing tests pass

### DIFF
--- a/docs/user-guide/concepts/tools/tools_overview.md
+++ b/docs/user-guide/concepts/tools/tools_overview.md
@@ -90,7 +90,7 @@ If a tool name contains hyphens, you can invoke the tool using underscores inste
 result = agent.tool.read_all(path="/path/to/file.txt")
 ```
 
-## Tool Execution
+## Tool Executors
 
 When models return multiple tool requests, you can control whether they execute concurrently or sequentially. Agents use concurrent execution by default, but you can specify sequential execution for cases where order matters:
 


docs/api-reference/tools.md
```diff
@@ -20,6 +20,16 @@
 ::: strands.tools.watcher
     options:
       heading_level: 2
+::: strands.tools.executors
```
zastrowm, line 23: Do we have an issue tracking auto-generating these?  If not could you create one?

Also, add the issue to the docs paper-cuts

docs/user-guide/concepts/tools/executors.md
```diff
@@ -0,0 +1,46 @@
...
+    tools=[weather_tool, time_tool]
+)
+# or simply Agent(tools=[weather_tool, time_tool])
+
+agent("What is the weather and time in New York?")
+```
+
+Assuming the model returns `weather_tool` and `time_tool` use requests, the `ConcurrentToolExecutor` will execute both concurrently.
+
+> Note: The model may decide to return one tool use request at a time. Under these circumstances, the tools will execute sequentially. Concurrency is only achieved if the model returns multiple tool use requests in a single response.
```
zastrowm, line 24: This is good, because we get a *lot* of questions about this.

Can we emphasize this even more:

 - Different models are better than others at parallel tool calling (e.g. Claude 4 is better than 3.7)
 - The system-prompt can be used to help force the LLM to invoke in parallel

Maybe even linking to documentation from others (like Anthropic: https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/implement-tool-use#parallel-tool-use)

mkdocs.yml
```diff
@@ -85,6 +85,7 @@ nav:
         - Overview: user-guide/concepts/tools/tools_overview.md
         - Python: user-guide/concepts/tools/python-tools.md
         - Model Context Protocol (MCP): user-guide/concepts/tools/mcp-tools.md
+        - Executors: user-guide/concepts/tools/executors.md
```
zastrowm, line 88: This is good if folks already know what executors are, but if they don't I'm not sure they would look here.  Is there a way we can make it more obvious when folks are reading about tools.  E.g. can we add a section to the tools_overview.md that actually calls out parallel vs sequential?

zastrowm (approved):

PR #304, opened 2025-10-30
Add titles to every readme page + add test to validate it

zastrowm:
## Description

I'm incrementally working towards getting our docs working with astro.build - to do so markdown files need a title front-matter.

Currently, [mkdocs does this](https://www.mkdocs.org/user-guide/writing-your-docs/#meta-data) to derive the title:

> The "title" to use for the document.
>
> MkDocs will attempt to determine the title of a document in the following ways, in order:
>
> 1. A title defined in the [nav](https://www.mkdocs.org/user-guide/configuration/#nav) configuration setting for a document.
> 2. A title defined in the `title` meta-data key of a document.
> 3. A level 1 Markdown header on the first line of the document body.
>    ([Setext-style](https://daringfireball.net/projects/markdown/syntax#header) headers are supported *only since MkDocs 1.5*.)
> 4. The filename of a document.
>
> Upon finding a title for a page, MkDoc does not continue checking any additional sources in the above list.

However, astro.build desires the title to be in the front-matter; for compatibility with both, add the title from the nav and add it as a frontmatter meta property.  Also added a test to enforce this for now

## Type of Change

- New content addition

## Motivation and Context

Incrementally let's docs work for both astro.build and mkdocs

## Checklist

- [X] Images/diagrams are properly sized and formatted

## Additional Notes

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -11,6 +11,19 @@ on:
 permissions: read-all
 
 jobs:
+  test:
+    name: Run Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          pip install .
+      - name: Run tests
+        run: |
+          pytest tests/ -v
+
   build:
     name: Build Documentation
     runs-on: ubuntu-latest
--- a/README.md
+++ b/README.md
@@ -71,6 +71,16 @@ mkdocs serve
 
 This will start a server at http://127.0.0.1:8000/ for previewing the documentation.
 
+### Running Tests
+
+To run all tests:
+
+```bash
+pytest tests/ -v
+```
+
+Tests are automatically run in CI/CD on pull requests and daily builds.
+
 ## Contributing ❤️
 
 We welcome contributions! See our [Contributing Guide](CONTRIBUTING.md) for details on:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,4 +12,5 @@ dependencies = [
     "mkdocstrings-python~=1.16.10",
     "mkdocs-llmstxt~=0.2.0",
     "strands-agents>=1.6.0",
+    "pytest>=7.0.0",
 ]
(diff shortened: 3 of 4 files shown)


.github/workflows/build-test.yaml
```diff
@@ -11,6 +11,19 @@ on:
...
 jobs:
+  test:
+    name: Run Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          pip install .
```
Unshure: nit: Do we need to setup python first? https://github.com/strands-agents/sdk-python/blob/main/.github/workflows/test-lint.yml#L59

docs/api-reference/multiagent.md
```diff
@@ -1,3 +1,6 @@
+---
+title: Multiagent
```
Unshure:
```suggestion
title: Multi-agent
```

docs/community/model-providers/fireworksai.md
```diff
@@ -1,3 +1,6 @@
+---
+title: Fireworks AI
+---
 # FireworksAI
```
Unshure: Should there be a space here or in the above title?

docs/community/session-managers/agentcore-memory.md
```diff
@@ -1,3 +1,6 @@
+---
+title: Amazon AgentCore Memory
```
Unshure: Does it matter if this doesnt match the below top-level header?

docs/api-reference/multiagent.md
zastrowm: I will commit to do this as a follow-up PR, but want to keep this PR scoped to using the titles as configured previously.

.github/workflows/build-test.yaml
zastrowm: Apparently not LOL:

https://github.com/strands-agents/docs/blob/de19eabdc76cc8b26e1864c9f5749dd127cb4739/.github/workflows/build-test.yaml#L18

zastrowm:
I misunderstood what title is doing and need to re-think the approach  here

PR #348, opened 2025-12-02
fix: Add Amazon Nova to Community Sidebar

zastrowm:
Quick follow up to #347

Add Amazon Nova to Community Sidebar + add community macro to the top of the page

-----


5herlocked:
@zastrowm -- it's not a community provider. This needs to be under just regular providers.

PR #483, opened 2026-01-27
docs: clarify short-term memory behavior in the memory example

azarboon:
## Description

**What changed:**  
This change clarifies that the memory used in the memory agent workflow is short-term, using the same terminology as the mem0 documentation, and is scoped to the current session by default. It also directs users to the relevant configuration and documentation for enabling persistence across sessions. Links to the appropriate mem0 documentation are added to improve discoverability and navigation.

**Why this change is needed:**  
The current documentation is largely silent and somewhat misleading regarding memory persistence. This ambiguity is compounded by the fact that mem0’s own documentation is vague on this topic. I had to spend time verifying the actual behavior, and this change aims to save that effort for future readers. The updated text aligns expectations with the default behavior and provides clear pointers to configuration options for enabling persistence.

## Type of Change

- New content
- Content update/revision

## Checklist

----

### DIFF
--- a/docs/examples/python/memory_agent.md
+++ b/docs/examples/python/memory_agent.md
@@ -18,7 +18,7 @@ The memory agent utilizes two primary tools:
 
 1. **memory**: Enables storing and retrieving information with capabilities for:
 
-    - Storing user-specific information persistently
+    - Storing user-specific information, either persistently or temporarily
     - Retrieving memories based on semantic relevance
     - Listing all stored memories for a user
     - Setting relevance thresholds and result limits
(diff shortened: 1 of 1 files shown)


ryanycoleman:
Hi @azarboon, thanks for the contribution! I did some digging into the mem0 default behavior before merging to make sure the note is accurate.

Looking at the example's source code ([`memory_agent.py`](https://github.com/strands-agents/docs/blob/main/docs/examples/python/memory_agent.py)), it uses `mem0_memory` from `strands_tools` with no explicit vector store configuration. By default, mem0 uses a local Qdrant backend that persists to `~/.mem0/` on disk so memories actually **survive across sessions** without any additional setup.

I verified this by tracing through the mem0 source:
- [`memory/setup.py`](https://github.com/mem0ai/mem0/blob/main/mem0/memory/setup.py) creates a `~/.mem0/` directory for persistent config and data
- [`memory/main.py`](https://github.com/mem0ai/mem0/blob/main/mem0/memory/main.py) initializes the vector store from defaults, which writes to that local directory
- The [mem0 docs on memory types](https://docs.mem0.ai/core-concepts/memory-types) describe both short-term and long-term memory, but the default local setup provides persistence

So the original wording ("Storing user-specific information persistently") appears to be accurate for the default config. The proposed note saying memory is "short-term and scoped to the current session" could mislead users into thinking they need extra setup for persistence, when persistence is actually the default.

That said, your instinct that users might want alternative options is helpful. Would you be open to revising the note to something like:

> By default, mem0 persists memories to local storage at `~/.mem0/`. See [mem0 vector store configuration](https://docs.mem0.ai/components/vectordbs/config) for options on using a different backend for production use.

_...and please accept our apologies for sitting on this PR!_

zastrowm (approved):

PR #491, opened 2026-01-28
Add a decision for opt-in breaking changes

zastrowm:
Add a record of the decision that "opt-in breaking changes" being acceptable from an SDK evolution strategy.

----


team/DECISIONS.md
```diff
@@ -99,3 +99,24 @@ The tradeoff is that `HookProvider` exposure can leak implementation details for
...
+
+### Decision
+
+Small breaking changes that follow the "pay for play" principle are acceptable without a major version bump. Programs can call new APIs to access new features, but programs that choose not to do so are unaffected — old code continues to work as it did before.
+
+### Rationale
+
+Strict semver adherence can slow SDK development when the breaking change only affects users who explicitly adopt new functionality. If existing code paths remain unaffected, the practical impact on users is minimal.
+
+For example, converting a `TypedDict` to `total=False` is technically breaking if existing implementations don't provide the new optional members. However, if the only way to encounter those new members is by adding a new tool that uses the new format, the change is effectively "pay for play." Users who don't adopt the new tool never observe the break.
```
mkmeral: Up until this point, I was just thinking "that's how backwards compatibility works" :P

But this makes sense. We have done something similar with Bidirectional Agents and tool context for example. We have started to pass in Bidi Agent as Agent. This is "technically" backwards compatible, but it's pay for play

pgrayy: It might just be me but I feel this explanation is slightly confusing.

> For example, converting a `TypedDict` to `total=False` is technically breaking if existing implementations don't provide the new optional members

I think I understand that you mean to say if a previously required field now becomes optional, existing code paths expecting the field would break if not provided. I just wonder if there is a more direct way to say this.

mehtarac (approved):

PR #497, opened 2026-01-28
Propose low-level snapshot api

zastrowm:
## Description

Unblock customers who would like to explicitly control snapshots of an agent.  Related issue: https://github.com/strands-agents/sdk-python/issues/1138

----

### DIFF
--- a/designs/snapshot-api.md
+++ b/designs/snapshot-api.md
@@ -1,0 +1,135 @@
+# Design Doc: Low-Level Snapshot API
+
+**Status**: Proposed
+
+**Date**: 2026-01-28
+
+**Issue**: https://github.com/strands-agents/sdk-python/issues/1138
+
+## Context
+
+Developers need a way to preserve and restore the exact state of an agent at a specific point in time. The existing SessionManagement doesn't address this:
+
+- SessionManager works in the background, incrementally recording messages rather than full state. This means it's not possible to restore to arbitrary points in time.
+- After a message is saved, there is no way to modify it and have it recorded in session-management, preventing more advance context-management strategies while being able to pause & restore agents.
+- There is no way to proactively trigger session-management (e.g., after modifying `agent.messages` or `agent.state` directly)
+
+## Decision
+
+Add a low-level, explicit snapshot API as an alternative to automatic session-management. This enables preserving the exact state of an agent at a specific point and restoring it later — useful for evaluation frameworks, custom session management, and checkpoint/restore workflows.
+
+### API Changes
+
+```python
+class Snapshot:
+    type: str              # the type of data stored (e.g., "agent")
+    state: dict[str, Any]  # opaque; do not modify — format subject to change
+    metadata: dict         # user-provided data to be stored with the snapshot
+
+class Agent:
+    def save_snapshot(self, metadata: dict | None = None) -> Snapshot:
+        """Capture the current agent state as a snapshot."""
+        ...
+
+    def load_snapshot(self, snapshot: Snapshot) -> None:
+        """Restore agent state from a snapshot."""
+        ...
+```
+
+### Behavior
+
+Snapshots capture **agent state** (data), not **runtime behavior** (code):
+
+- **Agent State** — Data persisted as part of session-management: conversation messages, context, and other JSON-serializable data. This is what snapshots save and restore.
+- **Runtime Behavior** — Configuration that defines how the agent operates: model, tools, ConversationManager, etc. These are *not* included in snapshots and must be set separately when creating or restoring an agent.
+
+The intent is that anything stored or restored by session-management would be stored in a snapshot — so this proposal is *not* documenting or changing what is persisted, but rather providing an explicit way to do what session-management does automatically.
+
+### Contract
+
+- **`metadata`** — Caller-owned. Strands does not read, modify, or manage this field. Use it to store checkpoint labels, timestamps, or any application-specific data.
+- **`type` and `state`** — Strands-owned. These fields are managed internally and should be treated as opaque. The format of `state` is subject to change; do not modify or depend on its structure.
+- **Serialization** — Strands guarantees that `type` and `state` will only contain JSON-serializable values.
+
+### Future Concerns
+
+- Snapshotting for MultiAgent constructs: This proposal would 
+- Providing a storage API for snapshot CRUD operations (save to disk, database, etc.)
+- Providing APIs to customize serialization formats
+
+## Developer Experience
+
+### Evaluations via Rewind and Replay
+
+```python
+agent = Agent(tools=[tool1, tool2])
+snapshot = agent.save_snapshot()
+
+result1 = agent("What is the weather?")
+
+agent2 = Agent(tools=[tool3, tool4])
+agent2.load_snapshot(snapshot)
+result2 = agent2("What is the weather?")
+# Compare result1 and result2
+```
+
+### Advanced Context Management
+
+```python
+agent = Agent(conversation_manager=CompactingConversationManager())
+snapshot = agent.save_snapshot(metadata={"checkpoint": "before_long_task"})
+
+# ... later ...
+later_agent = Agent(conversation_manager=CompactingConversationManager())
+later_agent.load_snapshot(snapshot)
+```
+
+### Persisting Snapshots
+
+```python
+import json
+from dataclasses import asdict
+
+agent = Agent(tools=[tool1, tool2])
+agent("Remember that my favorite color is orange.")
+
+# Save to file
+snapshot = agent.save_snapshot(metadata={"user_id": "123"})
+with open("checkpoint.json", "w") as f:
+    json.dump(asdict(snapshot), f)
+
+# Later, restore from file
+with open("checkpoint.json", "r") as f:
+    data = json.load(f)
+snapshot = Snapshot(**data)
+
+agent = Agent(tools=[tool1, tool2])
+agent.load_snapshot(snapshot)
+agent("What is my favorite color?")  # "Your favorite color is orange."
+```
+
+### Edge cases
+
+Restoring runtime behavior (e.g., tools) is explicitly not supported:
+
+```python
+agent1 = Agent(tools=[tool1, tool2])
+snapshot = agent1.save_snapshot()
+agent_no = Agent(snapshot)  # tools are NOT restored
+```
+
+## Consequences
+
+**Easier:**
+- Building evaluation frameworks with rewind/replay capabilities
+- Implementing custom session management strategies
+- Creating checkpoints during long-running agent tasks
+- Cloning agents (load the same snapshot into multiple agent instances)
+- Resetting agents to a known state (we do this manually for Graphs)
+
+**More difficult:**
+- N/A — this is an additive API
+
+## Willingness to Implement
+
+Yes
--- a/designs/snapshot-api.md
+++ b/designs/snapshot-api.md
@@ -6,6 +6,19 @@
 
 **Issue**: https://github.com/strands-agents/sdk-python/issues/1138
 
+## Motivation
+
+Today, developers who want to manually snapshot and restore agent state can *almost* do so by saving and loading these properties directly:
+
+- `Agent.messages` — the conversation history
+- `Agent.state` — custom application state
+- `Agent._interrupt_state` — internal state for interrupt handling
+- Conversation manager internal state — state held by the conversation manager (e.g., sliding window position)
+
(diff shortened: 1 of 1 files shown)


strands-agent:

**Preview URL**: https://d3ehv1nix5p99z.cloudfront.net/pr-497/

strands-agent:

**Preview URL**: https://d3ehv1nix5p99z.cloudfront.net/pr-497/

JackYPCOnline (dismissed):

designs/snapshot-api.md
```diff
@@ -0,0 +1,135 @@
...
+        ...
+
+    def load_snapshot(self, snapshot: Snapshot) -> None:
+        """Restore agent state from a snapshot."""
+        ...
+```
+
+### Behavior
+
+Snapshots capture **agent state** (data), not **runtime behavior** (code):
```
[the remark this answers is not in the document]
zastrowm: I view the tool state as a feature that we'd be adding to the agent to make "enabled" tools into a state on the agent.  So, if we had that I imagine it would be something like:

```python
agent = Agent(tools=[tool1, tool2, tool3, tool4], enabled_tools=["tool1"])
```

Where only `tool1` would be enabled/available on the agent.  Then to enable other tools something would eventually trigger:

```python
agent.enabled_tools = ["tool1", "tool3"]
```

and for restoring an agent with specific tools, it would be the same as 

```python
agent2 = Agent(tools=[tool3, tool4])
agent2.load_snapshot(snapshot)
```

and the snapshot would be restoring the `enabled_tools` state back into the agent.

designs/snapshot-api.md
```diff
@@ -0,0 +1,135 @@
...
+```python
+agent = Agent(tools=[tool1, tool2])
+snapshot = agent.save_snapshot()
+
+result1 = agent("What is the weather?")
+
+agent2 = Agent(tools=[tool3, tool4])
+agent2.load_snapshot(snapshot)
+result2 = agent2("What is the weather?")
+# Compare result1 and result2
```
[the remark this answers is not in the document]
zastrowm: I'll update it to better illustrate that it's evaluating the result of result1 and result2. 

> and maybe do we want to enforce tools?
> 
> ...if you restore a list of messages with a toolset of (1,2) to an agent with toolset of (3,4); you are a lot more likely to get hallucinations.

ACK that this is a concern, but IMO this is not the goal of the snapshot api. Snapshots are intended to only save/load the agent state - transformation of state or normalizing would be something that could be built on top of the low level primitive.  If I "resume" an agent from a snapshot or a session-management, it shouldn't be doing any conversion/munging of behavior on the way in or out.

designs/snapshot-api.md
```diff
@@ -0,0 +1,174 @@
...
+
+### API Changes
+
+```python
+class Snapshot(TypedDict):
+    type: str              # the type of data stored (e.g., "agent")
+    state: dict[str, Any]  # opaque; do not modify — format subject to change
+    metadata: dict         # user-provided data to be stored with the snapshot
+
+class Agent:
```
mkmeral: Do we want specific agent methods, or `Snapshottable` (wip name) interface? 

Having an interface like below, would make implementation a lot easier. And it would allow us to extend to other types (multi-agent, etc)  

```
Snapshottable:
   def save():
   def load():
```

This is similar to @JackYPCOnline 's idea on `SessionAware`but different naming pretty much

designs/snapshot-api.md
```diff
@@ -0,0 +1,174 @@
...
+    def save_snapshot(self, metadata: dict | None = None) -> Snapshot:
+        """Capture the current agent state as a snapshot."""
+        ...
+
+    def load_snapshot(self, snapshot: Snapshot) -> None:
+        """Restore agent state from a snapshot."""
+        ...
```
afarntrog: Great idea! I would consider moving these methods to their own class and inject it in the agent class `agent.snapshot.load()`

designs/snapshot-api.md
```diff
@@ -0,0 +1,174 @@
...
+Snapshots capture **agent state** (data), not **runtime behavior** (code):
+
+- **Agent State** — Data persisted as part of session-management: conversation messages, context, and other JSON-serializable data. This is what snapshots save and restore.
+- **Runtime Behavior** — Configuration that defines how the agent operates: model, tools, ConversationManager, etc. These are *not* included in snapshots and must be set separately when creating or restoring an agent.
+
+The intent is that anything stored or restored by session-management would be stored in a snapshot — so this proposal is *not* documenting or changing what is persisted, but rather providing an explicit way to do what session-management does automatically.
+
+### Contract
+
+- **`metadata`** — Application-owned. Strands does not read, modify, or manage this field. Use it to store checkpoint labels, timestamps, or any application-specific data without the need for a separate/standalone object/datastore.
```
dbschmigelski: nit: naming is hard but I think describing this as metadata is a stretch. If it is just `store checkpoint labels, timestamps` I agree but the `any application-specific data` part seems like this is broader

designs/snapshot-api.md
```diff
@@ -0,0 +1,174 @@
...
+```python
+agent = Agent(tools=[tool1, tool2])
+snapshot = agent.save_snapshot()
+
+result1 = agent("What is the weather?")
+
+# ...
+
+agent2 = Agent(tools=[tool3, tool4])
+agent2.load_snapshot(snapshot)
```
afarntrog: We should consider allowing for passing in the snapshot in the Agent init (as well)

designs/snapshot-api.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Decision
+
+Add a low-level, explicit snapshot API as an alternative to automatic session-management. This enables preserving the exact state of an agent at a specific point and restoring it later — useful for evaluation frameworks, custom session management, and checkpoint/restore workflows.
+
+### API Changes
+
+```python
+class Snapshot(TypedDict):
+    type: str              # the type of data stored (e.g., "agent")
```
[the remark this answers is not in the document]
poshinchen: +1, Probably it can be in the sate / metadata.

I notice that otel trace / span properties are good to have in many cases. I wonder if some of them can be added into state or metadata.

designs/snapshot-api.md
```diff
@@ -0,0 +1,174 @@
...
+## Up for Debate
+
+### What state should be included in a snapshot?
+
+The current proposal includes:
+
+- **messages** — conversation history
+- **interrupt state** — internal state for paused/resumed interrupts
+- **agent state** — custom application state (`agent.state`)
+- **conversation manager state** — internal state of the conversation manager (but not the conversation manager itself)
```
Unshure: We dont need this for the proposed implementation here. This was what bridged the gap to restoring an agent if we only persist the ongoing transcript of messages

designs/snapshot-api.md
```diff
@@ -0,0 +1,174 @@
...
+        """Restore agent state from a snapshot."""
+        ...
+```
+
+### Behavior
+
+Snapshots capture **agent state** (data), not **runtime behavior** (code):
+
+- **Agent State** — Data persisted as part of session-management: conversation messages, context, and other JSON-serializable data. This is what snapshots save and restore.
+- **Runtime Behavior** — Configuration that defines how the agent operates: model, tools, ConversationManager, etc. These are *not* included in snapshots and must be set separately when creating or restoring an agent.
```
pgrayy: So does that mean a configuration like `model_id` wouldn't be stored in the snapshot? Is there a specific reason why? May have missed it.

designs/snapshot-api.md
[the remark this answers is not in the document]
afarntrog: +1 the term "snapshot" makes me think of a disk snapshot - literally everything. I would like to see this incorporate tools etc in the future.

designs/snapshot-api.md
```diff
@@ -0,0 +1,174 @@
...
+- **agent state** — custom application state (`agent.state`)
+- **conversation manager state** — internal state of the conversation manager (but not the conversation manager itself)
+
+This draws a distinction between "evolving state" (data that changes as the agent runs) and "agent definition" (configuration that defines what the agent *is*):
+
+| Evolving State (snapshotted) | Agent Definition (not snapshotted) |
+|------------------------------|-----------------------------------|
+| messages | system_prompt |
+| interrupt state | tools |
+| agent state | model |
```
dbschmigelski: note for if we do model, we should be persisting model configuration probably be a ModelSnapshot not an str

franktip:
I previously implemented a checkpointing feature (not generally available unfortunately) and have a few general comments about the proposed feature based on my experience:
  * it would be useful to specify exactly what is included in a "snapshot". From my experience, I think this should include at least the following: (i) the entire state of the agent, which should include the entire chat history that led to the current state, (ii) the LLM that was used, (iii) the tools that were used.  Note that (i) is necessary if you want the agent to resume from the snapshot because (setting context management aside) the next query to the LLM typically includes the entire prior history. 
      - It would be good to define an interface that represents the agent 's state and format used for a snapshot declaratively (e.g. using JSON) that analysis/visualization tools can use.
  * the terminology "snapshot" suggests that it identifies a single point that was reached during the agent's execution. However, since it needs to include the entire prior history as noted above, it actually reflects the entire sequence of steps that the user and agent took to reach a given state. Therefore, adopting terminology like a "trajectory" may be more appropriate. In principle, given a trajectory, one should be able to resume from any step/iteration in that trajectory by simply discarding anything that follows that step.
  * keeping track of the state of the project that the agent is working on is one of the most important prerequisites for implementing checkpointing. If you want an agent to resume from a given snapshot, you need to make sure that the project/environment that the agent was working with is in the state that it was at that time. In other words, an agent needs to maintain a mapping from its internal state to the state of the environment/project that it is working on, so that this mapping can be used to determine how the environment can be restored appropriately. It would be good to avoid hard-coding the specific representation of the project/environment in the Strands agent itself because different clients of Strands may work with different environments (e.g. not all clients may work with git, and a client that works with a database would need to be able to reset the state of the database when resuming from a given snapshot).

designs/snapshot-api.md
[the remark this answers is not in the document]
zastrowm: Yea Timestamp _could_ be in metadata/user provided.  That said, it might be common enough to include top-level.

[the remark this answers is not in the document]
zastrowm: Because we're not handling the persistence of the snapshots, I'm not sure hash here makes sense. That said, there's going to be a strong need to make snapshots easy to persist and this might be a good feature to include as part of that

zastrowm:
> * it would be useful to specify exactly what is included in a "snapshot". From my experience, I think this should include at least the following: (i) the entire state of the agent, which should include the entire chat history that led to the current state, (ii) the LLM that was used, (iii) the tools that were used.  Note that (i) is necessary if you want the agent to resume from the snapshot because (setting context management aside) the next query to the LLM typically includes the entire prior history.

The P0 here would be that from a snapshot, you could pick up exactly where you were at (assuming the agent was created in the same initial way that it was before).  So from that perspective I think we're aligned.

That said:

> the LLM that was used,

Is this the LLM that "is configured for the next call" or "what was used in previous turns"?  The team wants to explore including the "current LLM configured", but I think the "What was used in previous turns" would be more part of observability and not necessarily what's included here.

>   
>   * It would be good to define an interface that represents the agent 's state and format used for a snapshot declaratively (e.g. using JSON) that analysis/visualization tools can use.\

I'll have to look into this.  I wanted to include the SDK data as a transparent blob so that we can evolve it going forward, but I do understand the need for evaluation purposes.  If not a P0, I can see defining the data-structure as a P1

> * the terminology "snapshot" suggests that it identifies a single point that was reached during the agent's execution. However, since it needs to include the entire prior history as noted above, it actually reflects the entire sequence of steps that the user and agent took to reach a given state. Therefore, adopting terminology like a "trajectory" may be more appropriate. In principle, given a trajectory, one should be able to resume from any step/iteration in that trajectory by simply discarding anything that follows that step.

I'll have to look into this a bit more to see how it compares to other frameworks.  Personally I'm not familiar with `trajectory` in this sense and so this would have not been obvious to me

> * keeping track of the state of the project that the agent is working on is one of the most important prerequisites for implementing checkpointing. If you want an agent to resume from a given snapshot, you need to make sure that the project/environment that the agent was working with is in the state that it was at that time. In other words, an agent needs to maintain a mapping from its internal state to the state of the environment/project that it is working on, so that this mapping can be used to determine how the environment can be restored appropriately. It would be good to avoid hard-coding the specific representation of the project/environment in the Strands agent itself because different clients of Strands may work with different environments (e.g. not all clients may work with git, and a client that works with a database would need to be able to reset the state of the database when resuming from a given snapshot).

Yeah, @Unshure is actively exploring some of this; on top of that, I think snapshots need a way that callers/plugins can contribute to snapshots, which is something I'm hoping the POC will help prove out

PR #528, opened 2026-02-05
docs(designs): Add skills sdk design

mkmeral:
WIP


designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+            self._active_skill = None
+            return f"Deactivated skill: {skill_name}"
+
+    @hook
+    def _inject_skill_metadata(self, event: BeforeInvocationEvent):
+        """Append skill metadata to system prompt before each invocation."""
+        ...
+```
+
+The `@tool` and `@hook` decorators inside the plugin class auto-register with the agent when the plugin is passed to the Agent constructor. This is the DX we want: declare what you need, the plugin protocol handles the wiring.
```
labeveryday: The plugin class does not show how the Plugin protocol discovers and registers. Can you clarify the DX claim?

labeveryday (commented):
This will be a nice addition to Strands. Looking forward to the release.

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+The plugin loads skills in `__init__` and auto-registers everything it needs when passed to the Agent: a `skills` tool, hooks for system prompt management, and hooks for tool filtering. You don't wire anything up manually.
+
+### How it works
+
+The `SkillsPlugin` implements the `Plugin` protocol. During `__init__`, it loads all skills. When the plugin is passed to the Agent:
+
+1. Loads skill metadata (name, description, location) from each `SKILL.md`
+2. Registers a `skills` tool on the agent with `activate` and `deactivate` actions
+3. Registers a `BeforeInvocationEvent` hook that appends skill metadata to the system prompt
+4. When a skill with `allowed_tools` is activated, optionally removes non-allowed tools from the agent (keeping them in memory for restoration on deactivate)
```
zastrowm: Can we up-level this into an agent property?; e.g. `tools_filter` that when enabled, only sends the tools with those names?

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+## Decision
+
+### Developer experience
+
+```python
+from strands import Agent
+from strands.plugins import SkillsPlugin
+
+agent = Agent(
+    plugins=[SkillsPlugin(skills=["./skills/code-review", "./skills/documentation"])]
```
afarntrog: Would this accept any file path? What about URL's (hosted skills)?
Have we considered just taking in a json schema that has a config for skills?

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+# Switch to different skills
+skills_plugin.skills = ["./skills/documentation"]
+agent("Write API docs for this module")
+
+# Add more skills
+skills_plugin.skills = ["./skills/documentation", "./skills/code-review"]
+agent("Document and review this module")
+```
+
+### Tracking the active skill
```
Unshure: How does claude code do this today? This is the mental model devs will have, so we should just follow what they do

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+```
+
+**Module exports:**
+
+```python
+# Skill dataclass from top-level
+from strands import Skill
+
+# Plugin from plugins submodule
+from strands.plugins import SkillsPlugin
```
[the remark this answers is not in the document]
mkmeral: +1 that's fair. We might want to move everything under `strands.plugins.skills` then?

designs/skills-support.md
mkmeral: I'll double check

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+
+### Tool restrictions with `allowed_tools`
+
+Skills can optionally specify which tools they're allowed to use:
+
+```yaml
+---
+name: safe-analyzer
+description: Analyzes files without executing code
+allowed-tools: file_read
```
afarntrog: I like this feature and would like to see it as 'disallowed-tools' as well.

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+```python
+# The plugin writes to agent.state during lifecycle events
+agent.state["skills_plugin"] = {
+    "skills": [str(s.path) for s in self._loaded_skills if s.path],
+    "active_skill": self._active_skill.name if self._active_skill else None,
+    "filtered_tools": [t.name for t in self._filtered_tools] if self._filtered_tools else None,
+}
+```
+
+Since `SessionManager` already persists `agent.state`, no special plugin-aware logic is needed. When the session is restored, the plugin reads `agent.state["skills_plugin"]` during initialization and re-applies skill configuration, active skill, and tool filtering.
```
afarntrog: Great idea! Some things we need to consider is what happens when an agent is reloaded and the skill is no longer present in the reloaded agent. I've seen this happen with tools. First agent run has a bunch of tools, the next run that is loaded from SessionManager does not have those tools

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+3. Registers a `BeforeInvocationEvent` hook that appends skill metadata to the system prompt
+4. When a skill with `allowed_tools` is activated, optionally removes non-allowed tools from the agent (keeping them in memory for restoration on deactivate)
+
+The agent sees something like this in its system prompt:
+
+```
+## Available Skills
+
+You have access to specialized skills. When a task matches a skill's
+description, use the skills tool to activate it and read its full instructions.
```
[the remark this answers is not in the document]
zastrowm: The spec has a comment on system-prompt integration: https://github.com/agentskills/agentskills/tree/main/skills-ref#agent-prompt-integration

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+
+```yaml
+---
+name: safe-analyzer
+description: Analyzes files without executing code
+allowed-tools: file_read
+---
+```
+
+Tool filtering is opt-in. When a skill with `allowed_tools` is activated, the plugin removes non-allowed tools from the agent, keeping them in memory. When the skill is deactivated, the full tool set is restored. If a skill doesn't specify `allowed_tools`, all tools remain available.
```
kevmyung: The AgentSkills.io spec defines `allowed-tools` as "a space-delimited list of tools that are pre-approved to run." The example — `Bash(git:*) Bash(jq:*) Read` — uses a tool + command pattern syntax, which reads as a permission rule: "this skill is pre-approved to run git/jq commands and Read without user confirmation." The spec's security section reinforces this with "Confirmation: Ask users before running potentially dangerous operations," implying `allowed-tools` is the mechanism to skip that confirmation.

This design reinterprets the same field as a tool visibility filter — "remove all tools except these from the agent." That's a meaningfully different semantic: one grants permission, the other restricts access.

The spec does say "Experimental. Support for this field may vary between agent implementations," so this isn't a spec violation. But as skills get shared across the ecosystem, a skill authored with `allowed-tools: Bash(git:*) Bash(jq:*) Read` under the spec's semantics would expect those tools to be pre-approved for execution — the author wouldn't intend for every other tool on the agent to be removed. Running that skill under this design would strip out tools the skill never meant to affect.

If the goal is compatibility with the broader AgentSkills.io ecosystem, `allowed-tools` should probably match the spec's semantics (skip human-in-the-loop confirmation for listed tools). If the filtering behavior is wanted as a separate feature, a different field name (e.g., `tool-filter` or `active-tools`) would avoid the mismatch.

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+
+Detection works through the `skills` tool: when the agent activates a skill, it becomes the active skill. It stays active until the agent explicitly deactivates it.
+
+### Multiple active skills
+
+Only one skill can be active at a time. Activating a new skill while one is already active implicitly deactivates the previous one (restoring any filtered tools before applying the new skill's `allowed_tools`). This keeps the mental model simple and avoids ambiguity around conflicting `allowed_tools` sets.
+
+If a future use case requires multiple simultaneous active skills, we can extend the design then. Starting with single-active is the safer default.
+
+### Tool restrictions with `allowed_tools`
```
kevmyung: This makes sense for keeping the plugin focused. But leaving tool registration entirely to the developer means all tools any skill might need have to be in Agent(tools=[...]) upfront. For generic tools like file_read and shell that's fine — the set stays small regardless of skill count.

But for agents with purpose-specific tools (e.g., web_search, generate_chart, create_presentation), the tool count grows with the skill catalog, and all their schemas are sent to the LLM on every turn whether or not the relevant skill is active. Progressive disclosure covers the instruction layer but not the tool layer.

I think this points to a broader concern with the design's relationship to tools. The plugin stays out of tool registration but still enters tool management through allowed_tools filtering and activate/deactivate. If the plugin were purely an instruction loader with no tool interaction, it'd be clean and focused. But once it touches tool visibility, it needs a coherent model for how tools and skills relate.

designs/skills-support.md
kevmyung: The AgentSkills.io spec defines allowed-tools as "a space-delimited list of tools that are pre-approved to run." The example — Bash(git:*) Bash(jq:*) Read — uses a tool + command pattern syntax, which reads as a permission rule: "this skill is pre-approved to run git/jq commands and Read without user confirmation." The spec's security section reinforces this with "Confirmation: Ask users before running potentially dangerous operations," implying allowed-tools is the mechanism to skip that confirmation.

This design reinterprets the same field as a tool visibility filter — "remove all tools except these from the agent." That's a meaningfully different semantic: one grants permission, the other restricts access.

The spec does say "Experimental. Support for this field may vary between agent implementations," so this isn't a spec violation. But as skills get shared across the ecosystem, a skill authored with allowed-tools: Bash(git:*) Bash(jq:*) Read under the spec's semantics would expect those tools to be pre-approved for execution — the author wouldn't intend for every other tool on the agent to be removed. Running that skill under this design would strip out tools the skill never meant to affect.

If the goal is compatibility with the broader AgentSkills.io ecosystem, allowed-tools should probably match the spec's semantics (skip human-in-the-loop confirmation for listed tools). If the filtering behavior is wanted as a separate feature, a different field name (e.g., tool-filter or active-tools) would avoid the mismatch

kevmyung (commented):
A few thoughts on the tool integration side:

If the design wants to support tool-backed skills (beyond instruction-only), it might be worth considering an approach where skills and tools are more explicitly connected — rather than relying on allowed-tools filtering and upfront tool registration.

- **Meta-tool routing instead of direct tool calls.** Rather than registering all skill tools on the agent and filtering them, a meta-tool like `skill_executor(skill_name, tool_name, tool_input)` could route execution internally. The agent's tool list stays stable, and tool schemas don't need to be registered upfront — they can be progressively disclosed through the skill activation flow alongside instructions.

- **Explicit skill-tool binding.** Rather than relying on `allowed-tools` for implicit tool association, a programmatic binding (e.g., a decorator or registration API) would make the skill-tool relationship explicit and enforceable in code.

For reference, I've been exploring this pattern [here](https://medium.com/@revoir07/agent-skills-part-2-bridging-skills-with-production-tool-ecosystems-422e4a63fcad) — meta-tool routing with decorator-based skill-tool binding, supporting local `@tool` functions and MCP tools under the same progressive disclosure model. Happy to share more details if useful.

designs/skills-support.md
```diff
@@ -0,0 +1,450 @@
...
+    def _inject_skill_metadata(self, event: BeforeInvocationEvent):
+        """Append skill metadata to system prompt before each invocation."""
+        ...
+```
+
+The `@tool` and `@hook` decorators inside the plugin class auto-register with the agent when the plugin is passed to the Agent constructor. This is the DX we want: declare what you need, the plugin protocol handles the wiring.
+
+### Resources and script execution
+
+The `SkillsPlugin` deliberately does not register tools for reading skill resources or executing skill scripts. The AgentSkills.io spec defines `scripts/`, `references/`, and `assets/` directories, but how those are accessed and executed depends entirely on the agent's environment.
```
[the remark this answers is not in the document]
mkmeral: > the tool count grows with the skill catalog, and all their schemas are sent to the LLM on every turn whether or not the relevant skill is active

I definitely agree. this is where I see skills integration with MCP can actually be scalable. You can do this logic through skills in MCP and bundle tools/skills together for specific use cases. Then for application developer, all you'd need is adding mcp server with skills to your agent. The orchestration layer underneath would handle all of the complexity

PR #530, opened 2026-02-06
Plugins - High level SDK features

Unshure:
Proposal for a way to introduce high-level features that have behavioral impact to an agents execution.

### DIFF
--- a/designs/0001-plugins.md
+++ b/designs/0001-plugins.md
@@ -27,7 +27,7 @@ Anthropic skills represent one of the many ways to build an Agent, but with some
 
 Strands SDK has defined its **low-level primitives** in such a way that any of these "agent-building techniques" are possible to implement, but there are issues with the recommended path.
 
-The current recommended approach in the sdk to add new functionality is to either introduce a new initialization parameter to the agent, or create a `HookProvider` which introduces this functionality. We cannot reasonably add `n` number of new parameters to the Agent, so the best path forward is to start creating and vending new `HookProvider` in the sdk. 
+The current recommended approach in the sdk to add new functionality is to either introduce a new initialization parameter to the agent, or create a `HookProvider` which introduces this functionality. While we could add `n` number of new parameters to the Agent, the devex starts to suffer and it makes it harder for customers to "fall into the pit of success", so instead a better path forward is to start creating and vending new `HookProvider` in the sdk. 
 
 Hook Providers ultimately solve our problem since they are composable, provide the necessary functionality to integrate rich features throughout an agents execution, and have been battle tested by many of our existing **high-level abstractions**.
 
@@ -40,7 +40,6 @@ agent = Agent(hooks=[AnthropicSkills()])
 
 agent = Agent(plugins=[AnthropicSkills()])
 ```
-**Note**: `Extension` is another option, but I am just biased toward plugins. Either way, we need a new named term to represent the introduction of these features in the sdk.
 
 I go into more detail about [Plugins and Hooks](#plugins-and-hooks) below.
 
@@ -52,7 +51,7 @@ I go into more detail about [Plugins and Hooks](#plugins-and-hooks) below.
 We will provide an abstraction for customers to implement their own plugins. While we may ship some with the sdk, customers are welcome to develop their own as well.
 
 #### Composable
-Multiple `Plugins` can be applied to a single agent. We make no garantees around plugins not conflicting as to provide maximum flexibility over what they can do.
+Multiple `Plugins` can be applied to a single agent. We make no guarantees around plugins not conflicting as to provide maximum flexibility over what they can do.
 
 #### Sharable
 Similar to our guidance on **low-level primitives**, we will recommend that community members share their own plugins by vending their own packages through some distributions mechanism (pypi, npm, github, etc).
@@ -73,7 +72,7 @@ From the backwards compatibility perspective, we cannot just remove these featur
 
 That being said, some features have a large impact on a majority of agents, or only make sense if there is "one" instance of it, so giving them a top-level agent initialization parameter help distinguish them from the feature-set of other `Plugins`. This also helps signal that they are "mostly" cross-compatible with any other `Plugin`.
 
-A good example of this is the Session Manager, where persisting and restoring an agent is a common and useful feature, and can contain additional functionality of persisting the state of other plugins ([see Appendix](#persisting-the-state-of-plugins)). We can decide to uplevel a plugin to an Agent class attribute based on its use case.
+A good example of this is the Session Manager, where persisting and restoring an agent is a common and useful feature, and can contain additional functionality of persisting the state of other plugins ([see Appendix](#persisting-the-state-of-plugins)). We can decide to up-level a plugin to an Agent class attribute based on its use case.
 
 
 ## Proposed SDK Changes
@@ -84,7 +83,7 @@ Add a `plugins` agent init parameter, and deprecate `hooks`.
 
 class Plugin(Protocol):
     name: str
-    def init_plugin(self, agent: Agent):
+    async def init_plugin(self, agent: Agent):
         ...
 
 agent = Agent(
@@ -121,7 +120,7 @@ A `hook` registered on an agent will invoke a callback function at a specified l
 # Plugin example
 class LoggingPlugin(Plugin):
     # Protocol method for a Plugin
-    def init_plugin(self, agent: Agent) -> None:
+    async def init_plugin(self, agent: Agent) -> None:
         agent.add_hook(BeforeInvocationEvent, self.log_start)
 
     def log_start(self, event: BeforeInvocationEvent) -> None:
@@ -139,16 +138,22 @@ agent.add_hook(BeforeInvocationEvent, log_start)
 
 While both of these examples attach a hook to the agent, there is room for improvement to meet our [We are accessible to humans and agents](https://github.com/strands-agents/sdk-python/blob/main/CONTRIBUTING.md?plain=1#L46) tenet where we say, "We don’t take shortcuts on curated DX for humans".
 
-To help improve this devex, a new `@hook_callback` decorator can be introduced to make it easy to wrap a function as a plugin, and then pass it into agent initialization. This aligns more closely with our "curated DX for humans" that the Strands team attempts to excel in.
+To help improve this devex, a new `@hook` decorator can be introduced to make it easy to wrap a function as a plugin, and then pass it into agent initialization. This aligns more closely with our "curated DX for humans" that the Strands team attempts to excel in.
 
 ```python
-@hook_callback
+@hook
 def log_calls(event: BeforeModelCallEvent): 
     print(f"Calling model...")
 
 agent = Agent(plugins=[log_calls])
 ```
-**Note**: `hook_callback` naming isn't decided upon yet, but the issue presented here is still valid. We should introduce an easy devex for authoring a simple `hook` callback. 
+**Note**: `hook` naming isn't decided upon yet, but the issue presented here is still valid. We should introduce an easy devex for authoring a simple `hook` callback. 
+
+### Follow-up items
+- Revisit and align on ["Simple `hook` callback authoring"](#simple-hook-callback-authoring)
+- Explore how persistence and state will be handled for `Plugins`
+- Build out and improve the `Plugin` protocol (like including convenience methods for adding hooks to the agent)
+- Provide a mechanism to order how `Plugins` are applied: https://github.com/strands-agents/sdk-python/issues/1593
 
 # Appendix
 
@@ -164,7 +169,7 @@ These features have related responsibilities, but their differences are nuanced,
 - **Plugin**: A high-level abstraction of the Strands SDK. A plugin represents some change to the standard behavior of an Agent, modifying the agents core attributes in order to elicit a desired behavior.
 - **Hooks**: A low-level primitive of the Strands SDK. A hook is a mechanism to execute code at a specific lifecycle event in the SDK, and gives relevant context at that specific lifecycle event in order to change the standard agents behavior.
 
-`Plugins` basically represent the usage/modification of **low-level primitives** like: system prompt, messages, tools, model, and **hooks**. The problem with this definition of `Plugins` is that it overlaps with the `HookProvider` abstraction currently present in the SDK. Well defined abstractions with clear separation of concerns align with the tenant [**The obvious path is the happy path**](https://github.com/strands-agents/sdk-python/blob/main/CONTRIBUTING.md?plain=1#L45), but including both of these goes against this tenant and can lead to customer confusion on which abstraction to choose. The code example below helps highlight this issue:
+`Plugins` basically represent the usage/modification of **low-level primitives** like: system prompt, messages, tools, model, and **hooks**. The problem with this definition of `Plugins` is that it overlaps with the `HookProvider` abstraction currently present in the SDK. Well defined abstractions with clear separation of concerns align with the tenet [**The obvious path is the happy path**](https://github.com/strands-agents/sdk-python/blob/main/CONTRIBUTING.md?plain=1#L45), but including both of these goes against this tenet and can lead to customer confusion on which abstraction to choose. The code example below helps highlight this issue:
 
 ```python
 # HookProvider example
@@ -186,7 +191,7 @@ agent = Agent(hooks=[LoggingHook()])
 # Note: Shape here is an example, not finalized
 class LoggingPlugin(Plugin):
     # Protocol method for a Plugin
-    def init_plugin(self, agent: Agent) -> None:
+    async def init_plugin(self, agent: Agent) -> None:
         agent.add_hook(BeforeInvocationEvent, self.log_start)
         agent.add_hook(AfterInvocationEvent, self.log_end)
 


designs/0007-plugins.md
```diff
@@ -0,0 +1,168 @@
...
+
+The current recommended approach in the sdk to add new functionality is to either introduce a new initialization parameter to the agent, or create a `HookProvider` which introduces this functionality. We cannot reasonably add `n` number of new parameters to the Agent, so the best path forward is to start creating and vending new `HookProvider` in the sdk. 
+
+Hook Providers ultimately solve our problem since they are composable, provide the necessary functionality to integrate rich features throughout an agents execution, and have been battle tested by many of our existing **high-level abstractions**.
+
+One problem with `HookProviders` is that their name does not best communicate what they are positioned to do. Their name overlaps with the **low-level primitives** hooks, so it can be not-so intuitive when applying a high-level feature, like Anthropic Skills, through a `hooks` parameter (see code below). I propose that we create a new `Plugin` concept, and a `plugins` agent init parameter, as plugins are a common term in the industry to represent adding new functionality to a system.
```
mkmeral: nit: skills is also cross-component. it'd be hooks, system prompt, tools, etc.

designs/0007-plugins.md
```diff
@@ -0,0 +1,168 @@
...
+
+That being said, some features have a large impact on a majority of agents, so giving them a top-level agent initialization parameter help distinguish them from the feature-set of other `Plugins`. This also helps signal that they are "mostly" cross-compatible with any other `Plugin`.
+
+A good example of this is the Session Manager, where persisting and restoring an agent is a common and useful feature, and can contain additional functionality of persisting the state of other plugins ([see Appendix](#persisting-the-state-of-plugins)). We can decide to uplevel a plugin to an Agent class attribute based on its use case.
+
+### Hooks vs HookProvider vs Plugins
+These features all have similar responsibilities, and the differences between them are nuanced, so here are definitions to explain what they do:
+
+- **Hooks**: A low-level primitive of the Strands SDK. A hook is a mechanism to execute code at a specific lifecycle event in the SDK, and gives relevant context at that specific lifecycle event in order to change the standard agents behavior.
+- **HookProvider**: A provider of hooks, with some convenience methods to make it easy to apply hooks to an agent.
```
mkmeral: I see HookProvider as just an implementation detail. With the `@hook` decorator, that should just be sth behind the scenes

designs/0007-plugins.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+With this approach, we deprecate `HookProviders` and the top-level `hooks` parameter, introduce a new `Plugin` protocol and `plugins` agent init arg, and start vending Agent behavior changing features as `Plugins` (like steering).
+
+An easy way to introduce low-level hooks is still valuable, so we can introduce a `@hook` decorator where you can wrap a function as a simple plugin.
+
+# Appendix
+
+## Persisting the state of Plugins
+The Session Manager has the responsibility of persisting and restoring the state of an agent, along with any attached plugins. The recommended way to accomplish this for now would be to have plugins maintain their state in the `agent.state`. Then during initialization, session manager would restore `agent.state`, and each plugin could restore themselves in turn from `agent.state`.
```
mkmeral: can't we just have those support snapshottable interface? if they want to? we can just load them back

strands-agent:

**Preview URL**: https://d3ehv1nix5p99z.cloudfront.net/pr-530/

designs/0007-plugins.md
```diff
@@ -0,0 +1,179 @@
...
+
+From the backwards compatibility perspective, we cannot just remove these features and make turn them into plugins, but we can re-think how we categorize these features. If a feature needs to perform an action that cannot be accomplished with the existing **low-level primitives** offered by the sdk, then the feature would likely not be considered a plugin. OTEL, for example, is a feature that may need to operate behind the scenes of the existing primitives (creating traces around the execution of primitives), so it does not fit well into the `Plugin` terminology. However the rest of the features mentioned in the list above can be reasonably defined as a `Plugin`.
+
+That being said, some features have a large impact on a majority of agents, so giving them a top-level agent initialization parameter help distinguish them from the feature-set of other `Plugins`. This also helps signal that they are "mostly" cross-compatible with any other `Plugin`.
```
zastrowm: Two additional deciding factors:

 - When a plugin is meant to be exclusive - e.g. you can only have "one" - two session-managers doesn't make sense because they clobber eachother
 - When a plugin modifies existing behavior - e.g. the retry-strategy.  In that case a plugin doesn't make sense because how do you indicate that you're adjusting values vs being "additive"

designs/0007-plugins.md
```diff
@@ -0,0 +1,179 @@
...
+We will provide an abstraction for customers to implement their own plugins. While we may ship some with the sdk, customers are welcome to develop their own as well.
+
+#### Composable
+Multiple `Plugins` can be applied to a single agent. While we cannot guarantee that these plugins will not conflict with one another, it leads to maximum flexibility over what they can do.
+
+#### Sharable
+Similar to our guidance on **low-level primitives**, we will recommend that community members share their own plugins by vending their own packages through some distributions mechanism (pypi, npm, github, etc).
+
+
+### What isn't a `Plugin`?
```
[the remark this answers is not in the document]
yonib05: Got it. How do we handle it if a plugin overwrites the session manager the customer used, for example? Does this mean plugins will need to have cross-compatibility or certify they work with other things?

designs/0007-plugins.md
```diff
@@ -0,0 +1,179 @@
...
+
+```python
+# Simple hook can still be defined as a plugin with a decorator
+@hook
+def log_calls(event: BeforeModelCallEvent): 
+    print(f"Got event: {event}")
+
+# Usage
+agent = Agent(
+    plugins=[log_calls]
```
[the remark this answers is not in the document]
dbschmigelski: Exactly, if @hook is just creating a plugin, it should be @plugin. But, we know that the concept of a hook still exists so we need to figure out how that fits in cleanly.

designs/0007-plugins.md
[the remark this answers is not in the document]
Unshure: We can incrementally work toward this, but as a P0 I dont think we should build those guarantees in. I mention this in the Composable section:
```
Multiple `Plugins` can be applied to a single agent. While we cannot guarantee that these plugins will not conflict with one another, it leads to maximum flexibility over what they can do.
```

We can probably do better, but before implementing them, I think its fine to lean into this flexibility

designs/0007-plugins.md
Unshure: Agree with the first point. But for retry strategy, Im a bit more on the fence. I could see someone adding an `AnthropicRetryStrategy` and an `OpenAIRetryStrategy` to the same agent through a plugin.

strands-agent:

**Preview URL**: https://d3ehv1nix5p99z.cloudfront.net/pr-530/

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+One problem with `HookProviders` is that their name does not best communicate what they are positioned to do. Their name overlaps with the **low-level primitives** hooks, so it can be not-so intuitive when applying a high-level feature, like Anthropic Skills, through a `hooks` parameter (see code below). I propose that we create a new `Plugin` concept, and a `plugins` agent init parameter, as plugins are a common term in the industry to represent adding new functionality to a system.
+
+```python
+# Between these two examples, I think the plugins example
+# does a better job communicating the feature
+agent = Agent(hooks=[AnthropicSkills()])
+
+agent = Agent(plugins=[AnthropicSkills()])
+```
+**Note**: `Extension` is another option, but I am just biased toward plugins. Either way, we need a new named term to represent the introduction of these features in the sdk.
```
zastrowm: Worth double-checking with product about naming in case they have a preference.

dbschmigelski (dismissed):

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+
+### What is a `Plugin`?
+
+`Plugins` provide a mechanism for customers to apply a behavior change (like Anthropic Skills) to an agent in a way that is extensible, composable, and shareable.
+
+#### Extensible
+We will provide an abstraction for customers to implement their own plugins. While we may ship some with the sdk, customers are welcome to develop their own as well.
+
+#### Composable
+Multiple `Plugins` can be applied to a single agent. While we cannot guarantee that these plugins will not conflict with one another, it leads to maximum flexibility over what they can do.
```
zastrowm:
```suggestion
Multiple `Plugins` can be applied to a single agent. We make no garantees around plugins not conflicting as to provide maximum flexibility over what they can do.
```

I'd almost say preventing incompatibility  an anti-goal, but a good goal would be allowing plugins the ability to query what other plugins are available.  This is used in Astro/Starlight for not adding functionality when a known plugin is installed, for example

zastrowm: I think this also goes into "what's a plugin" vs "what's a parameter" later - e.g. when we identify an area as important enough or is something where only a single "Component" should adjust, that's when it's no longer a plugin but a higher level concept

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+Today Strands ships with many **high-level abstractions** that the team has determined to be valuable to a majority of customers:
+- Session Manager
+- OTEL
+- Conversation Manager
+- RetryStrategy
+- Steering
+
+All of these features have been built into the sdk before the introduction of the concept of `Plugins`, but if they were built today, would potentially have been implemented as a `Plugin`. So this begs the question - What is or isn't a `Plugin`?
+
+From the backwards compatibility perspective, we cannot just remove these features and make turn them into plugins, but we can re-think how we categorize these features. If a feature needs to perform an action that cannot be accomplished with the existing **low-level primitives** offered by the sdk, then the feature would likely not be considered a plugin. OTEL, for example, is a feature that may need to operate behind the scenes of the existing primitives (creating traces around the execution of primitives), so it does not fit well into the `Plugin` terminology. However the rest of the features mentioned in the list above can be reasonably defined as a `Plugin`.
```
mehtarac: is renaming still not backwards incompatible?

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Proposed SDK Changes
+
+Add a `plugins` agent init parameter, and deprecate `hooks`.
+
+```python
+
+class Plugin(Protocol):
+    name: str
+    def init_plugin(self, agent: Agent):
```
zastrowm: Probably makes sense to be async - specifically for session-manager purposes? (I'm thinking TS for example)

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+A good example of this is the Session Manager, where persisting and restoring an agent is a common and useful feature, and can contain additional functionality of persisting the state of other plugins ([see Appendix](#persisting-the-state-of-plugins)). We can decide to uplevel a plugin to an Agent class attribute based on its use case.
+
+
+## Proposed SDK Changes
+
+Add a `plugins` agent init parameter, and deprecate `hooks`.
+
+```python
+
+class Plugin(Protocol):
```
mkmeral: is this the final protocol?

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+# After
+agent.add_hook(BeforeModelCallEvent, callback)
+```
+
+#### Simple `hook` callback authoring
+A `hook` registered on an agent will invoke a callback function at a specified lifecycle event. While a user can simply define an agent, then use the `agent.add_hook` method to attach the hook to the agent, there isn't a simple way to create attach a `hook` during agent initialization. Lets looks at a simple logging use case:
+
+```python
+# Plugin example
+class LoggingPlugin(Plugin):
```
mkmeral: are hooks plugins? 

can I do

```python

@hook
def myhook(event: ...):
   ...

agent(plugins=[myhook])
```

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+def log_start(self, event: BeforeInvocationEvent) -> None:
+    print(f"Request started for agent: {event.agent.name}")
+
+agent = Agent()
+agent.add_hook(BeforeInvocationEvent, log_start)
+```
+
+While both of these examples attach a hook to the agent, there is room for improvement to meet our [We are accessible to humans and agents](https://github.com/strands-agents/sdk-python/blob/main/CONTRIBUTING.md?plain=1#L46) tenet where we say, "We don’t take shortcuts on curated DX for humans".
+
+To help improve this devex, a new `@hook_callback` decorator can be introduced to make it easy to wrap a function as a plugin, and then pass it into agent initialization. This aligns more closely with our "curated DX for humans" that the Strands team attempts to excel in.
```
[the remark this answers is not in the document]
zastrowm: meh - not sure it really matters that much

Maybe my larger point is - do we need a streamed line way of creating a plugin instead?  Like a function that is called for all hooks? (callback_handler).

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+
+```python
+@hook_callback
+def log_calls(event: BeforeModelCallEvent): 
+    print(f"Calling model...")
+
+agent = Agent(plugins=[log_calls])
```
pgrayy: So is this saying that a plugin could be a Plugin instance or a hook callback?

designs/0007-plugins.md
Unshure: This is the rough shape. I dont want to do implementation review here, this is just meant to get agreement on roughly what a plugin does.

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+    def log_end(self, event: AfterInvocationEvent) -> None:
+        print(f"Request completed for agent: {event.agent.name}")
+
+# Passed in via the hooks parameter
+agent = Agent(plugins=[LoggingPlugin()])
+```
+
+The usage of each abstraction is essentially the same, so in order to avoid the customer confusion of which to choose, I'm proposing that we deprecate `HookProvider` in favor of `Plugin` moving forward. There is a world where `Plugins` can depend on `HookProviders`, and extend the behavior, but I think this leads to more indirection rather than proper abstractions. 
+
+To be clear, we are not deprecating the concepts of `hooks` in the SDK, which is the **low-level primitive** for running code at lifecycle events of an agent. We are just renaming the previous abstraction of `HookProvider` to `Plugin` to better communicate its intended functionality to customers.
```
JackYPCOnline: Maybe move it to the top? it is a foundimental idea

designs/0007-plugins.md
mkmeral: nvm saw it later :p

zastrowm (dismissed):

designs/0007-plugins.md
awsarron: let's dig in to that when appropriate. The current interface is very thin which raises some questions for me

designs/0007-plugins.md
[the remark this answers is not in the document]
Unshure:
```
from strands import hook

class LoggerPlugin(Plugin):

  @hook
  def logger(event: BeforeModelCallEvent)
  ..

@hook
def logger(event: BeforeModelCallEvent | AfterModelCallEvent):
  print(event)
```

This is our current best idea, but we are worried that the `@hook` decorator does not exactly imply that this is a plugin.

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Adding a new concept to Strands - Plugins
+
+Strands SDK has defined its **low-level primitives** in such a way that any of these "agent-building techniques" are possible to implement, but there are issues with the recommended path.
+
+The current recommended approach in the sdk to add new functionality is to either introduce a new initialization parameter to the agent, or create a `HookProvider` which introduces this functionality. We cannot reasonably add `n` number of new parameters to the Agent, so the best path forward is to start creating and vending new `HookProvider` in the sdk. 
+
+Hook Providers ultimately solve our problem since they are composable, provide the necessary functionality to integrate rich features throughout an agents execution, and have been battle tested by many of our existing **high-level abstractions**.
```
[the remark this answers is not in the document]
Unshure: These are mentioned earlier, but steering or session manager are good examples.

strands-agent:

**Preview URL**: https://d3ehv1nix5p99z.cloudfront.net/pr-530/

designs/0007-plugins.md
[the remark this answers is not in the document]
Unshure: We talked about having the ability to add hooks easily to the plugin:
```
class LoggerPlugin(Plugin):
  @hook
  def log(event: BeforeModelCallEvent):
    print(event)
```

But agree, we will explore this more during implementation.

designs/0007-plugins.md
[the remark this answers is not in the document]
Unshure: Renaming this to @hook for now, we will continue to explore what the best devex is here.

designs/0007-plugins.md
```diff
@@ -0,0 +1,205 @@
...
+agent = Agent()
+agent.add_hook(BeforeInvocationEvent, log_start)
+```
+
+While both of these examples attach a hook to the agent, there is room for improvement to meet our [We are accessible to humans and agents](https://github.com/strands-agents/sdk-python/blob/main/CONTRIBUTING.md?plain=1#L46) tenet where we say, "We don’t take shortcuts on curated DX for humans".
+
+To help improve this devex, a new `@hook_callback` decorator can be introduced to make it easy to wrap a function as a plugin, and then pass it into agent initialization. This aligns more closely with our "curated DX for humans" that the Strands team attempts to excel in.
+
+```python
+@hook_callback
```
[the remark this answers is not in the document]
Unshure: Going with @hook for now, but its not decided on yet. Will iterate before including.

pgrayy (approved):

afarntrog (approved):

Issue #441, opened 2026-01-15
Task: Migrate to new CMS


zastrowm:
We have reached a limit to where mkdocs is not scaling with us. We need a more robust system to help us manage docs and to that extent we'll be migrating to a new CMS/static-site generator.

### Tentative Plan

- https://github.com/strands-agents/docs/issues/442
- https://github.com/strands-agents/docs/issues/507
- https://github.com/strands-agents/docs/issues/508
- [x] Bug Bash & validate new docs
- [ ] Release updated docs/site to public
- [ ] Copy as Markdown
- [x] Update site repository to use docs repository

PR #538, opened 2026-02-11
feat: Add new CMS documentation preview

zastrowm:
## Description

Add a new workflow which creates previews of the documentation using a workflow in this repository instead of the site repository.  This is the flow that the new CMS will be using going forward.

This is a revamp of #536 - the original workflow was flawed because workflows running against pull_request target don't have access to secrets, and thus it cannot deploy to S3.  Instead we need to use pull_request_target which then means we need to use environments and the workflow needs to work on the main branch.

The workflow is a port of the one in #536 but now following the same flow we use for integ tests:

 - Determine which environment to deploy to (github org members get auto deployed) using the shared action
 - Subsequent steps build & deploy to S3

An example of this flow exists at https://github.com/zastrowm/docs/pull/11

A follow up change will remove the old workflow from the `workstream/cms-migration` branch, which is no longer needed (since it doesn't work).

## Related Issues

https://github.com/strands-agents/docs/issues/441

## Type of Change

- Other (please describe): Workflow

## Checklist

----

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -1,0 +1,168 @@
+name: Deploy Docs Preview
+
+on:
+  pull_request_target:
+    branches:
+      # TODO once we merge the new CMS to main, adjust this
+      - 'workstream/cms-migration'
+    types: [opened, reopened, synchronize]
+
+concurrency:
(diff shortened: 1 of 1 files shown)


strands-agent:

**Preview URL**: https://d3ehv1nix5p99z.cloudfront.net/pr-538/

PR #541, opened 2026-02-11
feat: CMS - Switch to copying files instead of hard-linking

zastrowm:
## Description

**Note**: This is targeting the `workstream/cms-migration` branch.

Previously we used a hard-link to include files in starlight/astro so that mkdocs and starlight would continue to work together - however, this is now slowing down progress in generating api documentation (Python & TypeScript). So switch to doing a straight copy to the directory.

Changes in this PR:

 - Copy the relevant files from `/docs` -> `src/content/docs` which is where starlight needs them; previously we used a directory link
 - `update-docs.ts` to account for this; as noted before, this is a script that will be used to do the final migration and so the contents are not to be tightly reviewed unless you see something obvious.  The final output is the more important part.

## Related Issues

#441

## Type of Change

- Content update/revision

## Checklist

----


scripts/update-docs.ts
```diff
@@ -64,6 +65,64 @@ async function getAllMarkdownFiles(dir: string): Promise<string[]> {
...
+
+  async function walk(currentDir: string) {
+    const entries = await readdir(currentDir, { withFileTypes: true });
+    for (const entry of entries) {
+      const fullPath = join(currentDir, entry.name);
+      if (entry.isDirectory()) {
+        await walk(fullPath);
+      } else if (entry.name.endsWith(".py")) {
+        files.push(fullPath);
+      }
```
mehtarac: Can this be just one function? Looks like the only difference is in the `else if` statements that checks for `.py` or `.ts` extension.

PR #543, opened 2026-02-12
designs - typescript - graph

pgrayy:
## Type of Change

- New content
- Content update/revision
- Typo/formatting fix
- Bug fix
- Other (please describe):

## Checklist

- [ ] I have tested the documentation locally using `mkdocs serve`

----

### DIFF
--- a/designs/typescript-graph.md
+++ b/designs/typescript-graph.md
@@ -1,0 +1,351 @@
+# TypeScript SDK - Graph - Design
+
+## Overview
+
+This document covers the design of the Graph multi-agent orchestration pattern for the TypeScript SDK. It walks through the core interfaces (Status, Node, Graph, Events), shows usage examples, and discusses future considerations around state isolation and remote execution.
+
+Not covered in this document:
+
+- Swarm builds on the same core interfaces introduced here (Status, Node, streaming events) and should not require a separate design discussion
+- Telemetry, hooks, and interrupts are planned for multi-agent patterns but follow closely from existing SDK patterns and should not require a separate design discussion
+- Session management will be reviewed separately
+
+## Interfaces
+
+The Graph pattern introduces a new `src/multiagent/` module containing shared multi-agent abstractions and the Graph-specific implementation.
+
+```
+src/
+└── multiagent/
+    ├── index.ts   # Public exports
+    ├── base.ts    # MultiAgentBase, MultiAgentState
+    ├── status.ts  # Status enum
+    ├── nodes.ts   # Node, NodeConfig, NodeInput, NodeResult, NodeState, AgentNode, MultiAgentNode, FunctionNode
+    ├── graph.ts   # Graph, GraphBuilder, GraphEdge, GraphInput, GraphResult, GraphState, GraphConfig
+    └── events.ts  # Streaming events
+```
+
+> [!NOTE]
+> Modules can be split into subdirectories later if they become unwieldy.
+
+### Base
+
+```typescript
+class MultiAgentState {
+  readonly user: z.infer<typeof schema>
+  ...
+}
+```
+
+- Base state class shared across multi-agent patterns (Graph, Swarm, etc.)
+- `user`: mutable section for customer-defined state. Typed via Zod schema inference (`z.infer<typeof schema>`) so customers get compile-time type safety and runtime validation without writing generics manually. Customers update it directly (e.g., `state.user.counter += 1`).
+
+### Status
+
+```typescript
+enum Status {
+  PENDING = 'PENDING',
+  EXECUTING = 'EXECUTING',
+  COMPLETED = 'COMPLETED',
+  FAILED = 'FAILED',
+  CANCELLED = 'CANCELLED',
+}
+```
+
+- Shared across nodes and graphs to track execution lifecycle
+- `PENDING`: not yet started
+- `EXECUTING`: currently running
+- `COMPLETED`: finished successfully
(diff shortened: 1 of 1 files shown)


designs/typescript-graph.md
```diff
@@ -0,0 +1,351 @@
+# TypeScript SDK - Graph - Design
+
+## Overview
+
+This document covers the design of the Graph multi-agent orchestration pattern for the TypeScript SDK. It walks through the core interfaces (Status, Node, Graph, Events), shows usage examples, and discusses future considerations around state isolation and remote execution.
+
+Not covered in this document:
+
+- Swarm builds on the same core interfaces introduced here (Status, Node, streaming events) and should not require a separate design discussion
```
mkmeral: (outside scope of this doc) one thing i'd like to see is how we can actually get rid of the tool injection logic. I don't like that swarm is so intrusive. 

Probably we should switch to structured output

designs/typescript-graph.md
```diff
@@ -0,0 +1,351 @@
...
+  readonly results: Record<string, NodeResult>
+  readonly error?: Error
+  ...
+}
+```
+
+- Aggregate result from a graph execution
+
+```typescript
+class GraphState extends MultiAgentState {
```
mkmeral: again for cyclic graphs. what if one execution succeeded and one failed? 

this is pretty annoying for Python bc we had to think about backwards compatibility with the already established data types/classes

designs/typescript-graph.md
```diff
@@ -0,0 +1,351 @@
...
+
+- Input passed to a node at execution time
+- Entry point nodes receive the original task provided to the graph
+- Downstream nodes receive content blocks built from upstream dependency outputs
+
+```typescript
+class NodeResult {
+  readonly nodeId: string
+  readonly status: Status
+  readonly duration: number
```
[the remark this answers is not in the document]
pgrayy: Agreed. We did not do this in Python. There are multiple duration fields where some are measured in seconds, others in milliseconds. Here I figured seconds was a good standard.

designs/typescript-graph.md
```diff
@@ -0,0 +1,351 @@
...
+
+This idea has implications beyond the graph — the `Agent` class itself stores state fields like `messages` directly as instance members. That means two concurrent `agent.invoke()` calls on the same agent instance would clobber each other's conversation history. The same session-manager-keyed-by-ID pattern could apply at the agent level to make individual agents concurrency-safe, not just graphs.
+
+### Remote Node
+
+A remote node would execute on a different process or machine — wrapping an agent behind an API, an MCP server, or an A2A endpoint. It would serialize input, send it over the wire, and deserialize the response back into `ContentBlock[]`. The `Node` abstraction already supports this — a `RemoteNode` subclass would implement `_stream()` with network calls instead of local execution, and the graph engine wouldn't need to change since it calls `node.stream()` uniformly.
```
zastrowm: This differs from RemoteAgent in that it elevates the concept of remote execution outside of an agent?  I like the idea.

There's nothing preventing this today in Python is there?  Assuming we support it via FunctionNode?

designs/typescript-graph.md
```diff
@@ -0,0 +1,351 @@
...
+  executionCount: number
+  ...
+}
+```
+
+- Per-node execution context created by the orchestration engine before each node runs
+- `multiAgentState`: reference to the current `MultiAgentState` (e.g., `GraphState`, `SwarmState`) for reading execution metadata and updating user state
+
+```typescript
+interface NodeConfig {
```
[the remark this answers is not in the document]
pgrayy: We could create a MutliAgentNodeConfig or AgentNodeConfig if need be but right now I'm not seeing a use case. But they would extend NodeConfig if we decided on it.

designs/typescript-graph.md
```diff
@@ -0,0 +1,351 @@
...
+class MultiAgentNode extends Node {
+  constructor(id: string, multiAgent: MultiAgentBase, config?: NodeConfig)
+}
+```
+
+- Wraps a `MultiAgentBase` instance, enabling nested orchestration (e.g., a `Graph` as a node in another `Graph`)
+- Forwards input to the nested pattern and wraps its streaming events with the node's ID
+
+```typescript
+type NodeHandler = (input: NodeInput, state: MultiAgentState) => NodeHandlerResult | Promise<NodeHandlerResult> | AsyncGenerator<MultiAgentStreamEvent, NodeHandlerResult, undefined>
```
zastrowm: In general I think we should prefer objects as arguments instead of individual arguments:

```suggestion
type NodeHandler = ({input: NodeInput, state: MultiAgentState}) => NodeHandlerResult | Promise<NodeHandlerResult> | AsyncGenerator<MultiAgentStreamEvent, NodeHandlerResult, undefined>
```

Allows folks to pick and choose and allows us to extend in the future

designs/typescript-graph.md
```diff
@@ -0,0 +1,351 @@
...
+//   researcher → writer → reviewer
+//                  ↑          |
+//                  └──────────┘  (condition: not approved)
+//                             |
+//                       formatOutput  (condition: approved)
+//
+const graph = new GraphBuilder()
+  .addNode(researcher)
+  .addNode(writer)
+  .addNode(reviewer)
```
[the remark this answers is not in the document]
pgrayy: We wrap the Agent instance into an AgentNode instance on behalf of the customer.

PR #549, opened 2026-02-13
feat: Generate API Documentation Pages

zastrowm:
## Description

**Note**: This is targeting the `workstream/cms-migration` branch.

Generate API documentation for both Python & TypeScript.  The way we do this in Starlight/Astro is by:

 - Cloning the sdk repositories
 - Generating markdown files from the repositories
 - Pull those markdown files into the CMS and publish under `api/{language}/thing`

### Additions

 - Scripts to clone SDKs and generate markdown (clone-sdks.ts, api-generation-python.py, api-generation-typescript.ts) - these *should* be reviewed as they're here long-term
 - Dynamic sidebar generation from content collection (dynamic-sidebar.ts) 
 - Python groups by module path, TypeScript groups by type category

### Notes

 - Both use flat slugs (api/typescript/Agent) but sidebar groups by category or modules via frontmatter
 - Index pages that auto-render module/type lists via Astro components
 - There are still links from pages that should be going to api documentation, but do not (they're pointing to the right place); I have not yet addressed that as this PR focuses on the pages

### Previews:

 - Python: https://d3ehv1nix5p99z.cloudfront.net/pr-cms-549/api/python
 - TypeScript: https://d3ehv1nix5p99z.cloudfront.net/pr-cms-549/api/typescript

## Related Issues

#441

## Type of Change

- Content update/revision

## Checklist

----

### DIFF
--- a/src/dynamic-sidebar.ts
+++ b/src/dynamic-sidebar.ts
@@ -1,4 +1,5 @@
 import type { StarlightRouteData } from '@astrojs/starlight/route-data'
+import { pathWithBase } from './util/links'
 
 type SidebarEntry = StarlightRouteData['sidebar'][number]
 type SidebarGroup = Extract<SidebarEntry, { type: 'group' }>
@@ -94,7 +95,7 @@ export function buildPythonApiSidebar(docs: DocInfo[], currentSlug: string): Sid
       if (hasDoc && !hasChildren) {
         // Leaf node - create a link
         const doc = value.__doc
-        const href = `/${doc.id}/`
+        const href = pathWithBase(`/${doc.id}/`)
         const link: SidebarLink = {
           type: 'link',
           label: displayName,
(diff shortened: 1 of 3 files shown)


scripts/clone-sdks.ts
```diff
@@ -0,0 +1,63 @@
...
+import { existsSync } from 'fs'
+import { join } from 'path'
+
+interface RepoConfig {
+  name: string
+  url: string
+  path: string
+}
+
+const repos: RepoConfig[] = [
```
Unshure: Anything special that would need to be done here for the evals sdk to be added? I dont want to add it now, but wondering if its just as easy as adding another item to this list?

Unshure (approved):
not to be figured out in this pr, but I see when I navigate to https://d3ehv1nix5p99z.cloudfront.net/pr-cms-549/api/, there is an error page. Is this just a side effect of cloudfront? If users navigate to a page that doesnt exist, what is the expected behavior?

PR #551, opened 2026-02-16
design: Add 0002-isolated-state proposal

pgrayy:
## Description
Design that proposes extracting per-invocation mutable state from the Agent instance into an isolated state object. Today, the Agent stores conversation history, user state, metrics, and other mutable data as instance fields, which prevents concurrent usage and creates fragile state management patterns for any code that needs to reset or manage agent state. This design explores using a session manager with invocation keys to isolate state per invocation.

## Related Issues
N/A

## Type of Change
- New content

## Checklist

----

### DIFF
--- a/designs/0002-isolated-state.md
+++ b/designs/0002-isolated-state.md
@@ -136,6 +136,7 @@ class InvocationState:
     event_loop_metrics: EventLoopMetrics
     trace_span: trace_api.Span | None
     interrupt_state: _InterruptState
+    ...
 ```
 
 The agent instance would retain only configuration: model, tools, system prompt, hooks, callback handler, conversation manager, etc. In the future, configuration could also be extracted into its own isolated object to allow per-invocation overrides, but this document focuses on invocation state to highlight the core problem and start the discussion.
@@ -163,7 +164,7 @@ async def stream_async(self, prompt, *, invocation_key=None, **kwargs):
 
 Because each invocation would operate on its own state object, there would be no shared mutable state on the agent. The `threading.Lock` and `ConcurrencyException` would no longer be needed.
 
-### Default behavior could preserve backwards compatibility
+### Default behavior and backwards compatibility
 
 One idea is to introduce a default in-memory session manager. Each agent instance would get a default invocation key that is stable across calls:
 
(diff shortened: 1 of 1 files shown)


designs/0002-isolated-state.md
```diff
@@ -0,0 +1,305 @@
...
+    )
+
+asyncio.run(main())
+```
+
+Each key would get its own isolated messages, agent state, metrics, and trace span. No lock contention, no `ConcurrencyException`.
+
+### Multi-agent patterns could become simpler
+
+With isolated state, multi-agent patterns that reuse the same agent instance become possible. For example, graph nodes could share a single agent and rely on unique invocation keys per execution:
```
mkmeral: why aren't they possible today?

designs/0002-isolated-state.md
```diff
@@ -0,0 +1,305 @@
+# Isolated State
```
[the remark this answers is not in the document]
pgrayy: 1. Snapshot can help with simplifying the serializing and deserializing and so this proposal would then mainly target the concurrency issue.
2. Graph is used as an example of user experience. I could also link to tickets raised by customers that led us to setting up the concurrency runtime error.

designs/0002-isolated-state.md
```diff
@@ -0,0 +1,305 @@
...
+import asyncio
+from strands import Agent
+
+agent = Agent(system_prompt="You are a helpful assistant.")
+
+async def main():
+    # This raises ConcurrencyException on the second call
+    results = await asyncio.gather(
+        agent.invoke_async("Summarize the Python GIL"),
+        agent.invoke_async("Summarize the Rust borrow checker"),
```
afarntrog: Is this something we want to encourage? Would you say it's an anti-pattern? Should we encourage folks to just have a func that returns a new instance of their defined agent?

afarntrog: I see you address in it "workaround" section :)

designs/0002-isolated-state.md
```diff
@@ -0,0 +1,305 @@
...
+# Must create separate instances even though they're identical
+graph.add_node(Agent(**summarizer_config), node_id="summarize_a")
+graph.add_node(Agent(**summarizer_config), node_id="summarize_b")
+```
+
+This goes against the SDK's goal of building agents in just a few lines of code.
+
+### State reset is fragile
+
+Any code that needs to reset an agent to a clean state must manually reach into its internals and know which fields to clear. This is error-prone — if the agent gains new stateful fields in the future, every reset site must be updated or it silently leaks state between executions.
```
dbschmigelski: I think this section ignores the simpler answer where we simply make the property private and add methods to make operating on them easier. I'd want to see pros and cons

designs/0002-isolated-state.md
```diff
@@ -0,0 +1,305 @@
...
+- Serving multiple users/conversations from a single agent instance
+
+### What could become harder or change
+
+- Internal code that currently reads `self.messages` or `self.state` would need to be updated to read from the invocation state object
+  - For example, hook callbacks that receive the agent and access `agent.messages` would need to be adapted
+- Session manager becomes a required concept (though a default in-memory implementation could make it invisible for simple use cases)
+- The `threading.Lock` and `ConcurrencyException` would be removed, which means users who relied on the exception as a signal would need to adapt
+
+### Backwards compatibility is the biggest concern
```
JackYPCOnline: Yes!!

PR #673, opened 2026-03-17
fix: update TypeScript snippets for latest SDK types

pgrayy:
## Description

Combines the changes from #667 and #657, plus fixes all `typecheck:snippets` errors caused by recent SDK changes.

Updates TypeScript code snippets to align with the latest SDK type renames and API changes:

- `AgentData` renamed to `LocalAgent` across hooks, plugins, and conversation management examples
- `MultiAgentBase` renamed to `MultiAgent` in multi-agent hook examples
- `S3Storage` moved to `@strands-agents/sdk/session/s3-storage` subpath import
- Zod updated from v3 to v4 to match SDK peer dependency
- Added `@aws-sdk/client-s3` to test-snippets dependencies for proper type deduplication

`npm run typecheck:snippets` passes with zero errors after these changes.

## Related Issues

Resolves #667
Resolves #657

## Type of Change

- Content update/revision
- Bug fix

## Checklist

---

### DIFF
--- a/src/content/docs/user-guide/concepts/plugins/index.ts
+++ b/src/content/docs/user-guide/concepts/plugins/index.ts
@@ -1,13 +1,11 @@
 import { Agent, FunctionTool, Tool } from '@strands-agents/sdk'
-import type { Plugin } from '@strands-agents/sdk'
+import type { LocalAgent, Plugin } from '@strands-agents/sdk'
 import {
   BeforeToolCallEvent,
   AfterToolCallEvent,
   BeforeModelCallEvent,
   AfterModelCallEvent,
 } from '@strands-agents/sdk'
-
-type AgentData = Parameters<Plugin['initAgent']>[0]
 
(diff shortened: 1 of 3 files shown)


src/content/docs/user-guide/concepts/plugins/index.ts
```diff
@@ -1,12 +1,14 @@
...
-import type { AgentData, Plugin } from '@strands-agents/sdk'
+import type { Plugin } from '@strands-agents/sdk'
 import {
   BeforeToolCallEvent,
   AfterToolCallEvent,
   BeforeModelCallEvent,
   AfterModelCallEvent,
 } from '@strands-agents/sdk'
 
+type AgentData = Parameters<Plugin['initAgent']>[0]
```
zastrowm: This should be a fast follow to fix this because I think if callers need the type we should export it. Having callers need to do this seems counter intuitive.

PR #731, opened 2026-04-02
docs: add state machine design doc (0005)

pgrayy:
## Description

Proposes restructuring the Agent loop into discrete steps coordinated by an orchestrator. The design decomposes the loop into five layers (Clients, Steps, Middleware, Plugins, Orchestrators) while keeping the public API unchanged.

Covers capabilities enabled by this decomposition: cross-cutting middleware (e.g., guardrails), checkpointing for durable execution, sub-orchestration for swappable execution strategies, and isolated invocation state for concurrent use.

### DIFF
--- a/designs/0005-state-machine.md
+++ b/designs/0005-state-machine.md
@@ -33,7 +33,7 @@ interface AgentState {
   // Dependencies
   model: Model
   toolRegistry: ToolRegistry
-  systemPrompt?: SystemPrompt
+  systemPrompt: SystemPrompt
   tracer: Tracer
   meter: Meter
   pluginRegistry: PluginRegistry
@@ -44,7 +44,7 @@ interface AgentState {
   messages: Message[]
   metrics: AgentMetric[]
   traces: AgentTrace[]
-  interrupt?: InterruptState
+  interrupt: InterruptState
   app: StateStore  // user-facing key-value state
 
   // Intra-loop temporaries (step-to-step communication)


designs/0005-state-machine.md
```diff
@@ -0,0 +1,356 @@
...
+  app: StateStore  // user-facing key-value state
+
+  // Intra-loop temporaries (step-to-step communication)
+  lastModelResult?: StreamAggregatedResult
+  structuredOutputChoice?: ToolChoice
+  ...
+}
+```
+
+`AgentContext` holds read-only dependencies:
```
mkmeral: nit: AgentState and AgentContext feels a bit confusing to me. Also context is overloaded term in AI imho. When I say agent context, first thing I think about is messages array

designs/0005-state-machine.md
```diff
@@ -0,0 +1,356 @@
...
+      state.messages,
+      this._buildStreamOptions(ctx, state)
+    )
+    state.lastModelResult = result
+    return { type: 'model', stopReason: result.stopReason, message: result.message }
+  }
+}
+```
+
+**ToolStep**: runs a single tool, yields progress events, and returns the tool result.
```
mkmeral: can we bundle them, or create hierarchy? e.g. `ToolsExecutionStep` is a list of `ToolStep`

mkmeral: comment to myself: read the doc and then write comments lol (the whole thing is explained below)

designs/0005-state-machine.md
```diff
@@ -0,0 +1,356 @@
...
+}
+```
+
+Example: a rate limiter that throttles step execution.
+
+```typescript
+class RateLimiter implements Middleware {
+  constructor(private _maxPerSecond: number) {}
+
+  wrap(step: Step): Step {
```
pgrayy: Comment from @zastrowm 
> This presumes that we have all steps pre-defined on this interface; any-way that we can make it more generic - like the way that hooks are (e.g. anyone can define a hook)

> I'm thinking of something like:
```typescript
agent.addMiddleware(ModelInvocationState, async (ctxt, state, next) => {
    await this._acquireToken()
    yield* next(ctx, state)
})
```

designs/0005-state-machine.md
```diff
@@ -0,0 +1,356 @@
...
+There are two kinds:
+
+**Built-in middleware** ships with the SDK and is always present. It's configured through state or context at runtime. One possible way to manage built-in middleware is via decorator syntax (`@`) on step class methods, though the exact mechanism is an implementation detail. Examples:
+
+| Middleware | What it does |
+|-----------|-------------|
+| `@traced` | Creates a telemetry span around the step, records result or error |
+| `@retryable` | Retries the step on transient errors with configurable backoff |
+
+**Custom middleware** is user-provided via the `middleware` param on the Agent constructor. It implements the `Middleware` interface:
```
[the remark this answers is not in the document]
pgrayy: Middleware would be similar to a plugin that performs some action on before and after events for a specific step. It however is more flexible in that it can directly alter execution by transforming inputs and outputs. Using plugins, we would have to build logic into the steps themselves to react to transformations to inputs and outputs (think of event.cancelTool).

PR #298, opened 2025-10-27
feat(mcp): add documentation for ToolProvider

dbschmigelski:
## Description

Updates MCP tools documentation to reflect the new experimental managed integration approach from PR #MCP . Adds comprehensive coverage of automatic lifecycle management alongside existing manual context management. Restructures content with clear integration approaches section and enhanced tool management examples. Refactors code examples to use realistic AWS documentation MCP server tools with expected results. Updates best practices and troubleshooting guidance for both integration methods. Streamlines language and improves overall documentation readability.

## Type of Change
- New content addition

## Areas Affected

`user-guide/concepts/tools/mcp-tools/`

## Screenshots

image
image
image
image

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] My changes follow the project's documentation style
- [x] I have tested the documentation locally using `mkdocs serve`
- [x] Links in the documentation are valid and working
- [x] Images/diagrams are properly sized and formatted
- [x] All new and existing tests pass

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #758, opened 2026-04-14
design: add Proactive Context Compression proposal

opieter-aws:
## Description                                                                                                                                                                      
                                                                                                                                                                                      
Adds design document 0006: Token-Aware Context Management. This proposes a `ThresholdConversationManager` that wraps any inner conversation manager to add proactive compression, tool result externalization, and in-loop context management.                

This is a follow-up on the roadmap in [0003-context-management.md](https://github.com/strands-agents/docs/blob/main/designs/0003-context-management.md)                                                                                                                                                                                                                                                     
                                                                                                                                                                                      
## Related Issues          
                                                                                                                                                                                    
- [#555: Proactive Context Compression](https://github.com/strands-agents/sdk-python/issues/555)
- [#1296: Large Tool Result Externalization](https://github.com/strands-agents/sdk-python/issues/1296)
- [#298: In-Event-Loop Context Management](https://github.com/strands-agents/sdk-python/issues/298)
- [#1295: Context Limit Property on Model](https://github.com/strands-agents/sdk-python/issues/1295)
- [#790: Track agent.messages token size](https://github.com/strands-agents/sdk-typescript/pull/790)
- [#2170: Automatic Context Optimization strategy](https://github.com/strands-agents/sdk-python/issues/2170)
                                                                                             
## Type of Change                                                                                                                                                                                                                                                                                                                                                         
- [x] New content                                                                                                                                                                   
- [ ] Content update/revision                                                                                                                                                       
- [ ] Structure/organization improvement                                                                                                                                            
- [ ] Typo/formatting fix                                                                                                                                                           
- [ ] Bug fix                                                                                                                                                                       
- [ ] Other (please describe):                                                                                                                                                      
                                                                                                                                                                                      
## Checklist                                                                                                                                                                        
                                                                                                                                                                                      

designs/0006-token-aware-context-management.md
lizradway, line 1: ---

One thing worth thinking about: this design is TS-specific, and the Python `ConversationManager` API is shaped pretty differently (separate `reduce_context` and `apply_management` methods, exception is required, no wrapper pattern today). Totally fine for a TS design doc, but wanted to flag it early so we don't end up with two divergent architectures for the same feature.

I think it'd be helpful to either:
1. Scope this explicitly as TS-only and note that a parallel Python design will follow, or
2. Propose shared abstractions both SDKs can converge on (e.g., agree that the wrapper pattern is the direction for Python too, that `reduce_context` should decouple from requiring an exception, etc.)

The roadmap currently targets Python first since that's where most of the customer demand is, so my recommendation would be to land the Python design first and let it inform the TS implementation.

lizradway, line 1: Wanted to share some thoughts on the scope, but am happy to discuss if you see it differently, as I have some concerns with `ThresholdConversationManager` as a bag of unrelated entities.

I noticed `ThresholdConversationManager` bundles three roadmap items ([#555](https://github.com/strands-agents/sdk-python/issues/555), [#1296](https://github.com/strands-agents/sdk-python/issues/1296), [#298](https://github.com/strands-agents/sdk-python/issues/298)) that were originally scoped separately because they have pretty different risk profiles and use cases:

- Externalization (1296): pure cost reducer, no LLM calls, low risk, shippable today
- Proactive compression (555): involves LLM calls, has a cost break-even curve, needs careful tuning
- In-loop management (298): architectural change to when hooks fire

The nice thing is these don't actually need a wrapper to compose, they hook into different events independently:
- Externalization hooks `AfterToolCallEvent`
- Proactive compression hooks `BeforeModelCallEvent`
- In-loop management falls out naturally since both hooks already fire within the agent loop

Keeping them separate would make it easier to ship incrementally (land the easy wins first), iterate on compression without touching externalization, and let users opt into just what they need.

My suggestion would be to focus this design on proactive compression (#555) as a standalone piece. Once all three are stable and proven independently, we could always offer a convenience flag or preset (e.g., `contextManagement: "auto"`) that turns on sensible defaults for all of them — but as sugar on top of independently shippable pieces rather than the starting point.

What do you think?

designs/0006-token-aware-context-management.md
```diff
@@ -0,0 +1,351 @@
...
+  /**
+   * Compression threshold as a fraction of the context window (0.0 to 1.0).
+   * When estimated input tokens / contextWindowLimit exceeds this value,
+   * reduce() is called on the inner manager before the next model call.
+   * @default 0.7
+   */
+  threshold?: number
+
+  /** Tool result externalization config. When provided, enables externalization. */
+  externalization?: ToolResultExternalizationConfig
```
lizradway: Externalization being optional but proactive compression being mandatory means that a user who only wants externalization has to set threshold: 1.0 as a workaround. This to me suggests a code smell these should be separate concerns rather than bundled in one manager.

designs/0006-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+```typescript
+const model = new BedrockModel({
+  modelId: 'anthropic.claude-sonnet-4-20250514',
+  tokenEstimator: (messages) => myTiktokenCount(messages),
+})
+```
+
+See [Alternative #3](#3-token-estimation-via-tokenizer) for why we chose heuristics as the default over a tokenizer dependency.
+
+**Decouple `reduce()` from `ContextWindowOverflowError`.** The current `ConversationManagerReduceOptions` requires an `error` field. Proactive compression calls `reduce()` before an error occurs, so `error` must become optional:
```
zastrowm: With TS, we intentionally picked making error required because the the ConversationManager's **only** responsibility was "Reduce the context in an emergency situation".  Anything else was considered a feature of the specific provider.

designs/0006-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+
+However, `ConversationManager` is the established primitive for context management. Users looking for context overflow solutions will look at conversation managers, not plugins. The existing `SlidingWindowConversationManager` already does proactive management (trimming after each invocation via `AfterInvocationEvent`), and the SDK's version supports `per_turn` for before-model-call management. Proactive compression fits naturally into this pattern. A plugin would also require exposing `conversationManager` on the `LocalAgent` interface so it could call `reduce()`, adding public API surface for an internal concern.
+
+### 2. Adding Threshold Config Directly to Existing Managers
+
+Issue [#555](https://github.com/strands-agents/sdk-python/issues/555) proposes adding `compressionThreshold` directly to `ConversationManager`. This is more discoverable but forces every conversation manager implementation to handle proactive logic. A wrapping manager keeps the threshold concern in one place and composes with any inner manager, including custom ones.
```
zastrowm: I think the DevX of having to import two ConversationManagers outweighs the need to re-implement.

If we implement the threshold check in the abstract `ConversationManager` base class, we get better DevX without requiring too much work on new managers. E.g. `SlidingWindowConversationManager` could implement `reduceOnThreshold` which, if implemented (and `threshold` is passed in the constructor), would enable the proactive trimming:

```typescript
class SlidingWindowConversationManager extends ConversationManager {
  constructor(config?: SlidingWindowConversationManagerConfig) {
    super({ threshold: config?.threshold })
    // ...
  }

  reduceOnThreshold({ agent }: ConversationManagerReduceOptions): boolean {
    return this._reduceContext(agent.messages)
  }
}

// usage
new SlidingWindowConversationManager({ windowSize: 50, threshold: 0.7 })
```

The downside would be that custom managers that don't implement `reduceOnThreshold` are unaffected.

zastrowm: This does violate the "Composition over inheritance" but again, I think the devx is a bit better for the common case

designs/0006-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+### Terminology
+
+- **`inputTokens`**: the number of input tokens reported by the model provider after a model call. This includes all messages, system prompt, and tool definitions sent in that request. It does not include output tokens generated by the model.
+- **`outputTokens`**: the number of tokens the model generated in its response. These tokens become part of the next call's input when the assistant message is appended to `agent.messages`.
+- **`contextWindowLimit`**: the maximum token capacity for a model, shared between input and output. Configured by the user or looked up from a per-provider table.
+- **threshold check**: comparing estimated next-call input tokens against `contextWindowLimit`. The default threshold (0.7) implicitly reserves headroom for the model to generate output.
+
+## Decision
+
+We implement proactive context compression as a new `ConversationManager` that wraps an inner manager. The wrapper handles when to compress (threshold-based, before each model call). The inner manager handles how to compress (sliding window, summarization, or any custom strategy).
```
mkmeral: > that wraps an inner manager

I don't like this. Should conversation managers be updated to support composibility? Or do we have the same problem in TS?

designs/0006-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+        Agent->>Tool: execute tool
+        Tool-->>Agent: tool result (potentially large)
+        Agent->>Agent: append tool result message
+        Note over Agent: go to next cycle
+    else response is endTurn
+        Note over Agent: invocation complete
+    end
+```
+
+After the model call, the framework attaches `usage` (containing `inputTokens` and `outputTokens`) to the assistant message as metadata ([#2125](https://github.com/strands-agents/sdk-python/pull/2125)). This metadata is part of the message itself, so it survives session persistence and restore. At the start of the next cycle, the conversation manager reads it directly from `agent.messages`:
```
mkmeral: qq, do we have any strategy about how we expose these trimming + conversation management operations through streaming? What is the DevX and end customer experience here?

designs/0008-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+
+If `reduce()` does not bring the token count below the threshold (for example, because the inner manager cannot remove enough messages), the manager skips compression on subsequent checks until new messages change the token count. This prevents repeated futile compressions, which is especially important when the inner manager is `SummarizingConversationManager` and each `reduce()` call triggers an LLM summarization.
+
+### SDK Changes Required
+
+This feature depends on three changes to the SDK, in addition to the `ThresholdConversationManager` itself.
+
+**Context window limit on Model ([#1295](https://github.com/strands-agents/sdk-python/issues/1295)).** The threshold check needs to know the model's context window size. [#1295](https://github.com/strands-agents/sdk-python/issues/1295) adds an optional `contextWindowLimit` to `BaseModelConfig` with per-provider lookup tables. Users can always override:
```
mkmeral, line 143: do we separate out `contextWindowLimit` from other params? 

adding a context window limit to model provider in my mind should impact bedrock behavior. so this devx feels a bit confusing to me

designs/0006-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+**Context window limit on Model ([#1295](https://github.com/strands-agents/sdk-python/issues/1295)).** The threshold check needs to know the model's context window size. [#1295](https://github.com/strands-agents/sdk-python/issues/1295) adds an optional `contextWindowLimit` to `BaseModelConfig` with per-provider lookup tables. Users can always override:
+
+```typescript
+const model = new BedrockModel({
+  modelId: 'anthropic.claude-sonnet-4-20250514',
+  contextWindowLimit: 200_000,
+})
+```
+
+**Message metadata ([#2125](https://github.com/strands-agents/sdk-python/pull/2125)).** The token counting strategy reads `inputTokens` and `outputTokens` from the last assistant message's metadata. This is already merged in the Python SDK. The TypeScript SDK will need an equivalent implementation.
```
lizradway: Also merged in typescript as of : https://github.com/strands-agents/sdk-typescript/pull/815

designs/0006-proactive-context-compression.md
mkmeral: +1 for this approach as "short term approach"

Also did we think about how it would work in TS over plugins?

designs/0006-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+
+**Token estimation.** The `Model` base class exposes an `estimateTokens(messages: Message[])` method with a default heuristic implementation: `chars / 4` for text content and `chars / 2` for JSON content. The method is designed to be overridden. Individual model providers can supply more accurate implementations using native counting APIs. Users can also pass a custom estimator via the `tokenEstimator` config option:
```
lizradway: Any reason why we not use the more accurate token estimate as default from the model provider API when available? All model providers should have this enabled in TS; I think it was only since Python had so many model providers that not all had their own estimator.

designs/0006-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+- **`outputTokens`**: the number of tokens the model generated in its response. These tokens become part of the next call's input when the assistant message is appended to `agent.messages`.
+- **`contextWindowLimit`**: the maximum token capacity for a model, shared between input and output. Configured by the user or looked up from a per-provider table.
+- **threshold check**: comparing estimated next-call input tokens against `contextWindowLimit`. The default threshold (0.7) implicitly reserves headroom for the model to generate output.
+
+## Decision
+
+We implement proactive context compression as a new `ConversationManager` that wraps an inner manager. The wrapper handles when to compress (threshold-based, before each model call). The inner manager handles how to compress (sliding window, summarization, or any custom strategy).
+
+```typescript
+export interface ThresholdConversationManagerConfig {
```
pgrayy: I would be interested to see a proposal that incorporates this logic into the existing ConversationManager interface. This way customers don't have to deal with multiple imports to setup their conversation manager. All the logic is centralized.

designs/0006-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+
+### 3. Token Estimation via Tokenizer
+
+We considered using a proper tokenizer like `tiktoken` as the default for token estimation instead of the `chars / 4` and `chars / 2` heuristics. A tokenizer produces more accurate counts, especially for non-ASCII text and code. However, the estimation only applies to new messages added since the last model call (the delta). The known baseline from `inputTokens + outputTokens` handles the bulk of the count. It also adds an external dependency that must be installed separately. The heuristics are dependency-free, work identically in every language, and are accurate enough for threshold-based decisions where the default already reserves 30% headroom. Users who need higher precision can pass a custom `tokenEstimator` to their model config, and we can override `estimateTokens()` for individual providers with native counting APIs where available.
```
lizradway: I commented on this above before I got to this section, but per "_we can override `estimateTokens()` for individual providers with native counting APIs where available_", when will be be doing this? by default or opt in flag? all should be available in TS, python is a different story.

designs/0006-proactive-context-compression.md
```diff
@@ -0,0 +1,256 @@
...
+
+### What Becomes Harder or Requires Attention
+
+Users must set `contextWindowLimit` on their model config until we ship per-provider default lookup tables. The manager depends on accurate `inputTokens` and `outputTokens` from providers. If a provider reports incorrect usage, the threshold check will be off. The `chars / 4` and `chars / 2` estimation heuristics are approximate and may over- or under-trigger for non-English text or unusual content types. Users who need higher precision can provide a custom `tokenEstimator`. Custom `ConversationManager` implementations that assume `error` is always defined in `reduce()` will need to handle the `undefined` case. The TypeScript SDK needs an equivalent of [#2125](https://github.com/strands-agents/sdk-python/pull/2125) (message metadata) before this feature can ship.
```
lizradway: Shipped in https://github.com/strands-agents/sdk-typescript/pull/815

designs/0006-proactive-context-compression.md
zastrowm: We should probably surface streaming events and/or hooks though I don't think we do right now

designs/0008-proactive-context-compression.md
opieter-aws: The placement of contextWindowLimit is scoped under #1295 (https://github.com/strands-agents/sdk-python/issues/1295), which proposes adding it to `BaseModelConfig` with per-provider lookup tables that auto-populate from `modelId.` The model is the source of truth for its own context window size, and other features beyond compression (token estimation, cost tracking) may also need it. The manual override would be rare in practice once lookup tables ship. 

Happy to brainstorm if you have any suggestions, let's discuss on #1295

designs/0006-proactive-context-compression.md
opieter-aws: Updated in the new design version

designs/0006-proactive-context-compression.md
opieter-aws: Updated in the new design version

designs/0006-proactive-context-compression.md
opieter-aws: Updated in the new design version

designs/0006-proactive-context-compression.md
opieter-aws: This is the plan, I rephrased it to make it more clear

designs/0006-proactive-context-compression.md
opieter-aws: Created an issue for this: https://github.com/strands-agents/sdk-python/issues/2172

designs/0006-proactive-context-compression.md
opieter-aws: We will be doing this by default. I updated the text to make that clearer.

designs/0008-proactive-context-compression.md
```diff
@@ -0,0 +1,231 @@
...
+### Token Counting Strategy
+
+The base class reads `inputTokens` and `outputTokens` from the last assistant message's metadata, then estimates any new messages added after it:
+
+```
+lastAssistant = last assistant message in agent.messages
+usage         = lastAssistant.metadata.usage
+
+if usage is available:
+    knownBaseline = usage.inputTokens + usage.outputTokens
```
lizradway, line 121: nit: we should be using and maintaining [latestContextSize](https://github.com/strands-agents/sdk-typescript/pull/790) so this value is exposed to the user.

non blocking implementation detail though

designs/0008-proactive-context-compression.md
```diff
@@ -0,0 +1,231 @@
...
+  modelId: 'anthropic.claude-sonnet-4-20250514',
+  contextWindowLimit: 200_000,
+})
+```
+
+**Message metadata ([#2125](https://github.com/strands-agents/sdk-python/pull/2125), [#815](https://github.com/strands-agents/sdk-typescript/pull/815)).** The token counting strategy reads `inputTokens` and `outputTokens` from the last assistant message's metadata. This is merged in both the Python SDK ([#2125](https://github.com/strands-agents/sdk-python/pull/2125)) and the TypeScript SDK ([#815](https://github.com/strands-agents/sdk-typescript/pull/815)).
+
+**Token estimation.** The `Model` base class exposes an `estimateTokens(messages: Message[])` method. Each model provider should implement this using its native token counting API (for example, Anthropic's `count_tokens()` or OpenAI's `tiktoken`). All TypeScript SDK providers have access to native counting APIs, so provider-level implementations are the expected default. The base class provides a heuristic fallback (`chars / 4` for text, `chars / 2` for JSON) for providers that do not implement native counting. Users can also pass a custom estimator via the `tokenEstimator` config option to override any provider's implementation:
```
lizradway, line 154: nice!

designs/0008-proactive-context-compression.md
```diff
@@ -0,0 +1,231 @@
...
+
+```typescript
+import { Agent, SummarizingConversationManager } from '@strands-agents/sdk'
+
+const agent = new Agent({
+  model: new BedrockModel({
+    modelId: 'anthropic.claude-sonnet-4-20250514',
+    contextWindowLimit: 200_000,
+  }),
+  conversationManager: new SummarizingConversationManager({ threshold: 0.7 }),
```
lizradway, line 179: `proactiveCompressionThreshold` is a bit wordy, but I think we'll need some more specification more than just `threshold` for the user to understand what this signifies.

nonblocking, can iterate on this naming in implementation

lizradway: I really really like this devx overall though, I love having threshold as an opt in flag that just runs a contained function on the existing conversation managers: modular/simple/understandable.

lizradway (approved):

pgrayy (approved):

designs/0008-proactive-context-compression.md
```diff
@@ -0,0 +1,231 @@
...
+    if no new messages:
+        tokenCount = knownBaseline
+    else:
+        tokenCount = knownBaseline + estimateTokens(newMessages)
+else:
+    // Cold start or no metadata: estimate everything
+    tokenCount = estimateTokens(agent.messages)
+
+if tokenCount / contextWindowLimit >= threshold:
+    reduceOnThreshold()
```
pgrayy, line 132: Spoke offline, going to see if existing reduce method could be utilized for both proactive and reactive management. The tricky part is that reduce currently requires the error message be passed through, which of course won't exist when running proactive.

PR #762, opened 2026-04-16
design: add Context Offload plugin

opieter-aws:
## Description

Adds design document 0009: Context Offloader. This documents the `ContextOffloader` plugin that intercepts oversized tool results via `AfterToolCallEvent`, stores each content block individually to a pluggable storage backend with content type preserved, and replaces the in-context result with a truncated preview plus per-block storage references. Includes a built-in `retrieve_offloaded_content` tool (enabled by default) so the agent can fetch offloaded content on demand.

Implements token-based thresholds using the model's `countTokens` method, with three storage backends: `InMemoryStorage`, `FileStorage`, and `S3Storage`.

Python implementation: [sdk-python#2162](https://github.com/strands-agents/sdk-python/pull/2162). TypeScript implementation: [sdk-typescript#974](https://github.com/strands-agents/sdk-typescript/pull/974).

Part of the [context management roadmap](https://github.com/strands-agents/docs/blob/main/designs/0003-context-management.md).

## Related Issues

- [#1296: Large Tool Result Externalization](https://github.com/strands-agents/sdk-python/issues/1296)
- [#1678: Large Content Aliasing](https://github.com/strands-agents/sdk-python/issues/1678)

## Type of Change

- New content

## Checklist

---


designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,167 @@
...
+
+1. **Data loss.** The full output is discarded permanently. The agent loses the ability to reference or reason about the content.
+
+2. **Reactive timing.** The replacement only happens after the model has already rejected the request. The oversized result consumes context space, triggers an overflow, wastes a round-trip, and only then gets truncated.
+
+This design proposes intercepting large tool results at execution time, before they enter the conversation. The full output is persisted to disk, and the conversation receives a truncated preview with a reference to the artifact file.
+
+## Decision
+
+We implement tool result externalization as a `Plugin` that hooks `AfterToolCallEvent`. If a tool result exceeds a configurable size threshold, the plugin externalizes the full output and replaces the conversation content according to a configurable strategy.
```
lizradway: I'm curious why a plugin instead of built into the SDK?

opieter-aws: I propose a plugin here because to my understanding that's the SDK's general purpose mechanism for reusable and configurable hooks on agent lifecycle events. This keeps the feature opt-in and separate from the tool execution path.

Building it into the SDK makes sense if we want to offer it as a core feature and increase discoverability. 

Does it make sense to first release it as a plugin, and based on adoption and feedback promote it to a built-in SDK feature?

designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,167 @@
+# Tool Result Externalization
```
lizradway: Just wanted to note some key differences between this TS design and my current Python implementation:
- **Image/JSON/Content Block processing scope**: current Python impl includes JSON/image/content block externalization, obscuring images/content blocks in the externalized preview. 
- **Retrieval Tools**: current Python impl includes a retrieval tool so that the user is not broken/blocked if they do not include the proper tool for externalized info retrieval (ie. S3 tool / file tool / etc). Additionally this helps for aliasing https://github.com/strands-agents/sdk-python/issues/1678 since it does not expose internal implementation and can be treated as a black box.
- **Default storage**: Since the primary value gain for tool result externalization is context window optimization, the current python impl treats in-memory as the default storage strategy. I think the main gain of in-file storage is that the externalized artifact is persisted after the agent closes, given that long-running agents are exactly the use case where context pressure matters most, I don't think this proves useful.



I think it would be worth including these in the TS implementation, can add more note about why inline.

opieter-aws: Thanks for the callout, these are great optimization points! I'll add them to the design here so it reflects the implementation state

designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,184 @@
...
+The default is `InMemoryExternalizationStrategy` because the primary value of externalization is context window optimization, not artifact persistence. Long-running agents are exactly the use case where context pressure matters most, and they do not benefit from writing artifacts to disk. Users who need persistence can opt into `FileExternalizationStrategy` or `S3ExternalizationStrategy`.
+
+When a tool result exceeds `sizeThreshold`, the hook passes the content to the strategy. The strategy stores the full output and returns a reference. The plugin then replaces the original content with a truncated preview plus that reference.
+
+The replacement content looks like:
+
+```
+[Externalized: 125,432 chars | ref: mem://abc123]
+
+<first 4,000 characters as preview>
```
lizradway: Magic number -- in python impl I have this set as an optional configuration param and think that could be worthwhile given the user's context management rigor / strategy.

designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,184 @@
...
+    Note over Plugin: if result > sizeThreshold:<br/>persist full output to storage backend<br/>replace content with preview + reference
+    Agent->>Agent: append tool result message
+    Note over Agent: go to next cycle
+```
+
+The plugin accepts a `sizeThreshold` (default 10,000 characters) and a pluggable `ExternalizationStrategy` that controls where the full output is stored:
+
+```typescript
+export interface ExternalizationStrategy {
+  externalize(content: string, toolName: string): Promise<string> | string
```
lizradway: I think this needs to be updated to include retrieve.

lizradway: also worth considering if we want to include full CRUD (mainly delete) now, or if we should think about object cleanup at a later date.

designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,184 @@
...
+
+1. **Data loss.** The full output is discarded permanently. The agent loses the ability to reference or reason about the content.
+
+2. **Reactive timing.** The replacement only happens after the model has already rejected the request. The oversized result consumes context space, triggers an overflow, wastes a round-trip, and only then gets truncated.
+
+This design proposes intercepting large tool results at execution time, before they enter the conversation. The full output is persisted to a pluggable storage backend, and the conversation receives a truncated preview with a reference to the stored artifact.
+
+## Decision
+
+We implement tool result externalization as a `Plugin` that hooks `AfterToolCallEvent`. If a tool result exceeds a configurable size threshold, the plugin externalizes the full output and replaces the conversation content according to a configurable strategy.
```
zastrowm: We should consider an opt-out mechanism as well.  Perhaps via a tool meta-tag or something (which doesn't exist yet) or configuration in the plugin itself.

I could imagine implementors wanting to always specify that certain tools are not externalized, for instance

designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,184 @@
...
+export interface ExternalizationStrategy {
+  externalize(content: string, toolName: string): Promise<string> | string
+}
+```
+
+The SDK ships three built-in strategies:
+
+| Strategy | Behavior | Use case |
+|----------|----------|----------|
+| `InMemoryExternalizationStrategy` | Stores content in memory. Zero config, no filesystem side effects. | Default. Context window optimization without persistence. |
```
zastrowm: I think we should be using AgentState - Otherwise if the plug gets pulled and the agent restored, it's losing a lot of information.

Snapshots should include this IMO

designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,184 @@
...
+| Image | Replaced with a `[image: format, N bytes]` placeholder. Follows the `SlidingWindowConversationManager` pattern. |
+| Document | Replaced with a `[document: format, name, N bytes]` placeholder. |
+
+The plugin measures the combined character count of all text and JSON blocks. If the total exceeds `sizeThreshold`, it externalizes the text content and replaces image and document blocks with lightweight placeholders.
+
+### Retrieval Tool
+
+The plugin vends a built-in retrieval tool that the agent can call to fetch externalized content by reference. This avoids requiring the user to separately configure a file-reading or S3 tool, and keeps the storage backend opaque to the model. The retrieval tool also supports the aliasing use case ([#1678](https://github.com/strands-agents/sdk-python/issues/1678)) because the model never needs to know where or how the content is stored.
+
+This operates at tool execution time, before the result enters the conversation history. It prevents a single large result from consuming a disproportionate share of the context window. The full output is preserved in the configured storage backend for later retrieval by the agent or the user.
```
gautamsirdeshmukh: I understand how the tool result is stored externally, that makes sense. But when the model retrieves it, does the full content just go back into the conversation context? 

More concretely: what prevents a loop where the retrieval tool returns content that itself exceeds the threshold, gets re-externalized, and the model retrieves it again? It seems like this tool may need some form of bounded access (pagination, etc)... otherwise very-long tool results could just cause the same context overflow the design is trying to fix

designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,184 @@
...
+
+Instead of a plugin, externalization could be a per-tool config or an agent-level setting. This would be more discoverable but would not compose as cleanly with other plugins. The plugin pattern keeps externalization independent and opt-in.
+
+## Consequences
+
+### What Becomes Easier
+
+Large tool results no longer blow up the context window or get silently discarded. The full output is preserved in the configured storage backend while the conversation receives a compact preview. The built-in retrieval tool allows the agent to fetch the full output on demand without requiring the user to configure a separate file or S3 tool.
+
+### What Becomes Harder or Requires Attention
```
zastrowm: This is true for all the handlers; I think we should have a plan on how we *might* address this going forward (message-ids, toolUseIds, or something else).  As we start to move towards longer-lived agents, memory pressure would even become a concern for the InMemoryStrategy

designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,184 @@
...
+  ],
+})
+
+// S3 storage: persists artifacts to a bucket
+const agent = new Agent({
+  tools: [dataAnalysis, apiClient, fileProcessor],
+  plugins: [
+    new ToolResultExternalizationPlugin({
+      sizeThreshold: 10_000,
+      strategy: new S3ExternalizationStrategy({
```
JackYPCOnline: The ExternalizationStrategy interface and its three built-in implementations (InMemory, File, S3) closely mirror the SnapshotStorage abstraction we already ship in the TypeScript SDK's session module (src/session/storage.ts), which has FileStorage and S3Storage implementations with the same tiering.

Rather than introducing a parallel storage abstraction, I'd suggest building externalization on top of the existing SnapshotStorage interface — or extracting a shared ContentStorage interface that both session management and tool result externalization compose

designs/0007-tool-result-externalization.md
lizradway: Nice! Like the updates :)

designs/0007-tool-result-externalization.md
```diff
@@ -0,0 +1,184 @@
...
+| Text | Externalized: stored in the backend, replaced with a truncated preview. |
+| JSON | Serialized via `JSON.stringify` (or `json.dumps` in Python), then externalized alongside text. |
+| Image | Replaced with a `[image: format, N bytes]` placeholder. Follows the `SlidingWindowConversationManager` pattern. |
+| Document | Replaced with a `[document: format, name, N bytes]` placeholder. |
+
+The plugin measures the combined character count of all text and JSON blocks. If the total exceeds `sizeThreshold`, it externalizes the text content and replaces image and document blocks with lightweight placeholders.
+
+### Retrieval Tool
+
+The plugin vends a built-in retrieval tool that the agent can call to fetch externalized content by reference. This avoids requiring the user to separately configure a file-reading or S3 tool, and keeps the storage backend opaque to the model. The retrieval tool also supports the aliasing use case ([#1678](https://github.com/strands-agents/sdk-python/issues/1678)) because the model never needs to know where or how the content is stored.
```
mkmeral: What features does this tool expose? I mean, getting the content as is does not work, because we would hit the same threshold again, right?

designs/0007-tool-result-externalization.md
lizradway: plus one --@zastrowm and I discussed this in detail on Friday and think it is likely out of the scope for this task in particular, but would be worth considering in the future since this is non-breaking implementation detail and can be changed.

lizradway: Same goes for Python externalization: https://github.com/strands-agents/sdk-python/pull/2162

designs/0007-tool-result-externalization.md
JackYPCOnline: why we have  externalize here but retrival as a tool

designs/0007-tool-result-externalization.md
lizradway: Might be worth cutting an issue to explore

designs/0007-tool-result-externalization.md
lizradway: interesting idea -- can explore

designs/0007-tool-result-externalization.md
lizradway: retrieval is tool that calls `retrieve` method internally

designs/0007-tool-result-externalization.md
lizradway: it initially returns the preview, aiming for the included information to be sufficient. 

since the tool result is externalized compression does not erase tool result.

designs/0007-tool-result-externalization.md
lizradway: https://github.com/strands-agents/sdk-python/issues/2168

designs/0007-tool-result-externalization.md
lizradway: yes, tool is not externalized if the tool is the built in retrieval tool (checks tool name)

mkmeral (approved):

PR #763, opened 2026-04-16
docs(designs): interventions primitive

lizradway:
interventions primitive design proposal

### DIFF
--- a/designs/0007-intervention-primitive.md
+++ b/designs/0007-intervention-primitive.md
@@ -208,7 +208,7 @@ User: "Query the secrets database for all API keys"
 
 **Deny** short-circuits immediately. **Interrupt** pauses if no handler denied. **Transform** modifies content and continues. **Guide** accumulates across handlers.
 
-No single handler catches everything — the value is in composition. See [Appendix C](#appendix-c-coverage-matrix) for the full matrix.
+No single handler catches everything — the value is in composition. See [Appendix D](#appendix-d-coverage-matrix) for the full matrix.
 
 ### Interrupt: Human-in-the-Loop
 
--- a/designs/0007-intervention-primitive.md
+++ b/designs/0007-intervention-primitive.md
@@ -1,0 +1,454 @@
+# Intervention: A First-Class Agent Control Primitive
+
+## Table of Contents
+
+- [Problem](#problem)
+- [The Insight: Intervention Is the Primitive](#the-insight-intervention-is-the-primitive)
+- [Why Not Separate Plugins?](#why-not-separate-plugins)
+- [Proposed API: `Agent(interventions=[...])`](#proposed-api-agentinterventions)
+- [Composability](#composability)
+- [Working Demos](#working-demos)
+- [Development Plan](#development-plan)
+- Appendices: [A (Concrete Instances)](#appendix-a-concrete-instances) · [B (Interface Design Rationale)](#appendix-b-interface-design-rationale) · [C (Coverage Matrix)](#appendix-c-coverage-matrix) · [D (Userland Workaround)](#appendix-d-userland-workaround) · [E (Naming)](#appendix-e-naming-alternatives)
+
+<details>
+<summary><h2>Definitions</h2></summary>
+
+| Term | Definition |
+|------|-----------|
+| **Plugin** | A Strands extension that hooks into agent lifecycle events (`BeforeToolCallEvent`, `AfterModelCallEvent`, etc.) and mutates the event object directly. The current mechanism for all agent control layers. |
+| **Steering** | A Strands vended plugin that uses an LLM to evaluate tool calls and model responses, returning Proceed, Guide (cancel + retry with feedback), or Interrupt (pause for human input). Currently Python-only. |
+| **Galileo Agent Control** | A Strands community plugin for runtime governance via configurable rules. Ships as two plugins (`AgentControlPlugin` for deny, `AgentControlSteeringHandler` for guide) because the current plugin interface can't express both. |
+| **Datadog AI Guard** | A Strands community plugin that scans for prompt injection, jailbreaking, and data exfiltration at four lifecycle points. |
+| **Cedar** | An open-source authorization policy language by AWS. Evaluates Allow/Deny decisions against principals, actions, resources, and context. Sub-ms, deterministic, formally verifiable. |
+| **OPA (Open Policy Agent)** | A general-purpose policy engine using the Rego language. CNCF graduated project. The main alternative to Cedar for policy-based authorization. |
+| **`invocation_state`** | A dict passed to a Strands agent on every call. Flows through the entire lifecycle — hooks and tools can read it. Used to carry user identity, roles, and environment context. |
+| **`InterruptException`** | The Strands SDK's mechanism for pausing agent execution and requesting human input. Raised by `event.interrupt()`, caught by the agent loop, and surfaced to the caller. |
+
+</details>
+
+---
+
+## Problem
+
+Strands agents have multiple independent control layers — authorization, steering, content guardrails, operational governance — but no shared interface between them. Each is a standalone plugin with its own action vocabulary, its own hook registration, and its own audit log. This creates four concrete problems:
+
+1. **No short-circuiting.** If Cedar denies a tool call in sub-ms, there's no way to skip the LLM steering call that's about to spend 100ms+ arriving at the same conclusion. Both plugins fire independently on every tool call because the framework doesn't know they're answering related questions.
+
+2. **Fragmented action model.** Steering returns Proceed/Guide/Interrupt but has no concept of Deny. Authorization needs Deny but has no concept of Guide. Galileo needs both and had to ship as two separate plugins (`AgentControlPlugin` + `AgentControlSteeringHandler`) because no single plugin interface can express "block this" and "retry with feedback" at the same time.
+
+3. **No unified audit.** Cedar logs to its audit trail, steering logs wherever steering logs, Datadog logs to Datadog. When someone asks "why did the agent delete that record?", you're correlating three log streams with different formats and no shared request ID.
+
+4. **Ordering is an implementation detail.** `Agent(plugins=[cedar, steering, guardrails])` looks ordered, but execution depends on hook registration internals that can change between SDK versions. There's no contract that cedar runs before steering.
+
+This proposal elevates the shared structure behind these control layers into a first-class SDK primitive: **Intervention**.
+
+---
+
+## The Insight: Intervention Is the Primitive
+
+Several independent tools already control agent behavior at runtime — [steering](https://strandsagents.com/docs/user-guide/concepts/plugins/steering/), [Galileo Agent Control](https://strandsagents.com/docs/community/plugins/agent-control/), [Datadog AI Guard](https://strandsagents.com/docs/community/plugins/datadog-ai-guard/), with [Cedar](https://www.cedarpolicy.com/) and [OPA](https://www.openpolicyagent.org/) authorization planned. They fall into two categories: **operational guardrails** (Galileo, Datadog, content guardrails) that enforce rules about *what's happening* regardless of who's doing it, and **authorization** (Cedar, OPA) that enforces rules about *who's allowed to do what*.
+
+They all answer different questions — but they share the same mechanical structure. They all: **intercept** an agent event, **evaluate** against rules, **decide** (proceed, redirect, or block), and **log** the decision. This shared lifecycle is the primitive — **Intervention**. Each control layer is an instance. See [Appendix A](#appendix-a-concrete-instances) for a detailed breakdown of each.
+
+The primitive has four components:
+
+**Events** — Typed event subclasses (`BeforeToolCallEvent`, `AfterModelCallEvent`, etc.) carry relevant context. Handlers only receive events they registered for.
+
+**Action** — Four decisions:
+
+| Action | Meaning |
+|--------|---------|
+| **Proceed** | Allow |
+| **Deny** | Hard block, no retry |
+| **Guide** | Cancel + feedback for retry |
+| **Interrupt** | Pause for human input |
+
+**Deny** is new — steering today only has Proceed/Guide/Interrupt. Authorization needs a hard block that means "you are not allowed, period."
+
+**Evaluation Engine** — Each instance uses a different engine (Cedar policies, LLM judge, API call, regex). The primitive doesn't prescribe how you evaluate, only what you return. See [Appendix A](#appendix-a-concrete-instances) for details on each.
+
+**Audit Trail** — Every handler logs its decision into a unified stream.
+
+---
+
+## Why Not Separate Plugins?
+
+Plugins communicate by mutating the event object. The framework sees the side effect but doesn't know what the plugin decided or why:
+
+```typescript
+// Plugin: mutates the event directly
+beforeToolCall(event: BeforeToolCallEvent): void {
+    if (!this.isAuthorized(event)) {
+        event.cancelTool = "Access denied"  // deny? guide? interrupt? framework can't tell
+    }
+}
+```
+
+An intervention handler returns a typed decision instead. The framework owns what happens next:
+
+```typescript
+// Four possible decisions — the shared vocabulary across all handlers
+type InterventionAction = Proceed | Deny | Guide | Interrupt
+// Intervention handler: returns a decision, framework applies it
+async evaluate(event: BeforeToolCallEvent): Promise<InterventionAction> {
+    if (!this.isAuthorized(event)) {
+        return new Deny("User not authorized for this tool")
+    }
+}
+```
+
+That distinction — returning a decision vs. mutating the event — is what makes the rest possible:
+
+**1. Short-circuiting.** When Cedar denies a tool call via a plugin, it sets `event.cancel_tool`. Steering runs next, makes an LLM call (~100ms), and returns Proceed — for a tool that's already cancelled. Plugins can't see each other. With interventions, Cedar returns `Deny` and the framework stops the pipeline immediately. Steering never runs.
+
+**2. A shared action model.** A plugin can set `event.cancel_tool = "..."` for deny, or do nothing for allow. There's no way to express "Guide" (cancel + retry with feedback) or "Interrupt" (pause for human input) through the same mechanism. Steering works around this by being a special plugin type with its own action vocabulary. Galileo works around it by shipping as two plugins. The intervention interface gives every handler the same four actions without special-casing.
+
+**3. Unified audit.** When debugging "why was this tool call blocked?", there's no single place to look — each plugin logs on its own (if it logs at all). The intervention registry records every handler's decision — handler name, event type, action, reason, principal, tool — into one stream at the framework level, regardless of whether individual services also log on their side.
+
+**4. Explicit conflict resolution.** If Cedar allows and Galileo denies, what happens? Today: both run, both mutate the event, last write wins. The outcome depends on plugin registration order, which is implicit. The intervention registry makes conflict resolution explicit: Deny wins over everything, Guide accumulates, Interrupt pauses if nothing denied.
+
+Separate plugins work for independent concerns that don't interact. Control layers interact — they all gate the same tool calls, they all need to know what the others decided, and they all benefit from not running when an earlier layer already blocked. That's what makes this a framework primitive, not a plugin concern.
+
+---
+
+## Proposed API: `Agent(interventions=[...])`
+
+Today, each control layer is a standalone plugin with no shared interface, no ordering guarantees, and no unified audit log:
+
+```typescript
+const agent = new Agent({
+    tools: [queryDatabase, sendEmail],
+    plugins: [cedarPlugin, steeringPlugin],
+    // Cedar and steering fire independently — no way to skip steering when Cedar denies
+})
+```
+
+With interventions as a first-class parameter:
+
+```typescript
+const agent = new Agent({
+    tools: [queryDatabase, sendEmail],
+    interventions: [cedar, guardrails, steering],  // cheapest first
+})
+```
+
+**Why first-class?** The framework owns composition — ordering, short-circuiting, conflict resolution, and a unified audit log are all built in. Steering becomes one instance of `InterventionHandler`, not a special concept.
+
+**Backwards compatibility:** Since the intervention primitive is being built in TypeScript first, and the existing control layers (steering, Galileo Agent Control, Datadog AI Guard) only exist in Python, there is no migration path — these will be implemented from scratch as `InterventionHandler` instances in TypeScript. The existing Python plugins continue to work unchanged for Python users. When Strands Python 2.0 ships with WASM bindings, the TypeScript intervention handlers become available in Python, and the Python-only plugins can be retired.
+
+### The `InterventionHandler` Interface
+
+The interface is **event-driven** — handlers declare which lifecycle events they care about, and the framework only calls them for matching events:
+
+```typescript
+abstract class InterventionHandler {
+    abstract name: string;
+    abstract handles(): Set<typeof HookEvent>;  // e.g. BeforeToolCallEvent, AfterModelCallEvent
+    abstract evaluate(event: HookEvent): Promise<InterventionAction>;
+}
+```
+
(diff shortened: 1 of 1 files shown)


designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+
+---
+
+## Problem
+
+Strands agents have multiple independent control layers — authorization, steering, content guardrails, operational governance — but no shared interface between them. Each is a standalone plugin with its own action vocabulary, its own hook registration, and its own audit log. This creates four concrete problems:
+
+1. **No short-circuiting.** If Cedar denies a tool call in sub-ms, there's no way to skip the LLM steering call that's about to spend 100ms+ arriving at the same conclusion. Both plugins fire independently on every tool call because the framework doesn't know they're answering related questions.
+
+2. **Fragmented action model.** Steering returns Proceed/Guide/Interrupt but has no concept of Deny. Authorization needs Deny but has no concept of Guide. Galileo needs both and had to ship as two separate plugins (`AgentControlPlugin` + `AgentControlSteeringHandler`) because no single plugin interface can express "block this" and "retry with feedback" at the same time.
```
mkmeral: > because no single plugin interface can express "block this" and "retry with feedback" at the same time.

I am not sure if I understand this part. Like do we expect plugin to return two results? what is the blocker here?

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+
+3. **No unified audit.** Cedar logs to its audit trail, steering logs wherever steering logs, Datadog logs to Datadog. When someone asks "why did the agent delete that record?", you're correlating three log streams with different formats and no shared request ID.
```
mkmeral: OTEL?

designs/0007-intervention-primitive.md
lizradway: The current plug in approach retry versus deny logic relies on interpreting the natural language of the cancellation method:

```
// Hard deny — stop, don't retry
event.cancel = "Access denied: analyst cannot delete records"

// Soft guide — retry with better input  
event.cancel = "Email body is too short. Add more context."
```

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+async evaluate(event: BeforeToolCallEvent): Promise<InterventionAction> {
+    if (!this.isAuthorized(event)) {
+        return new Deny("User not authorized for this tool")
+    }
+}
+```
+
+That distinction — returning a decision vs. mutating the event — is what makes the rest possible:
+
+**1. Short-circuiting.** When Cedar denies a tool call via a plugin, it sets `event.cancel_tool`. Steering runs next, makes an LLM call (~100ms), and returns Proceed — for a tool that's already cancelled. Plugins can't see each other. With interventions, Cedar returns `Deny` and the framework stops the pipeline immediately. Steering never runs.
```
zastrowm: In theory plugins should detect this situation; but the primitive makes this much more clear

mkmeral: is this a bug then? like can we just not run hooks for the rest if `event.cancel_tool` is set?

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+
+**2. A shared action model.** A plugin can set `event.cancel_tool = "..."` for deny, or do nothing for allow. There's no way to express "Guide" (cancel + retry with feedback) or "Interrupt" (pause for human input) through the same mechanism. Steering works around this by being a special plugin type with its own action vocabulary. Galileo works around it by shipping as two plugins. The intervention interface gives every handler the same four actions without special-casing.
```
mkmeral: > There's no way to express "Guide" (cancel + retry with feedback) or "Interrupt" (pause for human input) through the same mechanism

doesn't steering use hooks under the hood. there has to be a mechanism right? 

guide is tool result with instructions, and interrupt is an interrupt exception, no?

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+
+**3. Unified audit.** When debugging "why was this tool call blocked?", there's no single place to look — each plugin logs on its own (if it logs at all). The intervention registry records every handler's decision — handler name, event type, action, reason, principal, tool — into one stream at the framework level, regardless of whether individual services also log on their side.
```
mkmeral: > there's no single place to look — each plugin logs on its own (if it logs at all).

should we allow plugins to emit OTEL events? how does a new primitive solve this problem?

designs/0007-intervention-primitive.md
lizradway: I suppose this could be reconstructed from traces, but I think if we decide to go with the Intervention primitive, a built in decision log would be nice sugar for ease of customer/security audit etc.

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+**Action** — Four decisions:
+
+| Action | Meaning |
+|--------|---------|
+| **Proceed** | Allow |
+| **Deny** | Hard block, no retry |
+| **Guide** | Cancel + feedback for retry |
+| **Interrupt** | Pause for human input |
+
+**Deny** is new — steering today only has Proceed/Guide/Interrupt. Authorization needs a hard block that means "you are not allowed, period."
```
JackYPCOnline: Are these strategies composable with each other?

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+### The `InterventionRegistry`
+
+The framework provides an `InterventionRegistry` that wires handlers into the Strands hook system. It registers one callback per event type, dispatches to all matching handlers in registration order, and applies conflict resolution:
+
+- **Deny** short-circuits immediately — remaining handlers never run
+- **Guide** accumulates across handlers — feedback from all handlers is concatenated, then the tool is cancelled with the combined guidance so the agent can retry
+- **Interrupt** maps to `event.interrupt()` — the SDK's native pause/resume mechanism
+- **Proceed** continues to the next handler
+
+Every decision is logged to a unified audit trail accessible via `agent.interventionAuditLog`.
```
zastrowm: Nit - in the PR we should iterate on this; it seems odd to have this top level;

`agent.interventions.auditLog` seems cleaner

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+```typescript
+const agent = new Agent({
+    tools: [queryDatabase, sendEmail],
+    interventions: [cedar, guardrails, steering],  // cheapest first
+})
+```
+
+**Why first-class?** The framework owns composition — ordering, short-circuiting, conflict resolution, and a unified audit log are all built in. Steering becomes one instance of `InterventionHandler`, not a special concept.
+
+**Backwards compatibility:** Since the intervention primitive is being built in TypeScript first, and the existing control layers (steering, Galileo Agent Control, Datadog AI Guard) only exist in Python, there is no migration path — these will be implemented from scratch as `InterventionHandler` instances in TypeScript. The existing Python plugins continue to work unchanged for Python users. When Strands Python 2.0 ships with WASM bindings, the TypeScript intervention handlers become available in Python, and the Python-only plugins can be retired.
```
mkmeral: Why are we not bringing this feature to Python 1.0?

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+
+**3. Unified audit.** When debugging "why was this tool call blocked?", there's no single place to look — each plugin logs on its own (if it logs at all). The intervention registry records every handler's decision — handler name, event type, action, reason, principal, tool — into one stream at the framework level, regardless of whether individual services also log on their side.
+
+**4. Explicit conflict resolution.** If Cedar allows and Galileo denies, what happens? Today: both run, both mutate the event, last write wins. The outcome depends on plugin registration order, which is implicit. The intervention registry makes conflict resolution explicit: Deny wins over everything, Guide accumulates, Interrupt pauses if nothing denied.
+
+Separate plugins work for independent concerns that don't interact. Control layers interact — they all gate the same tool calls, they all need to know what the others decided, and they all benefit from not running when an earlier layer already blocked. That's what makes this a framework primitive, not a plugin concern.
+
+---
+
+## Proposed API: `Agent(interventions=[...])`
```
mkmeral: nit: is `intervention` a good name?

mkmeral: what are our alternatives?

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+**Why first-class?** The framework owns composition — ordering, short-circuiting, conflict resolution, and a unified audit log are all built in. Steering becomes one instance of `InterventionHandler`, not a special concept.
+
+**Backwards compatibility:** Since the intervention primitive is being built in TypeScript first, and the existing control layers (steering, Galileo Agent Control, Datadog AI Guard) only exist in Python, there is no migration path — these will be implemented from scratch as `InterventionHandler` instances in TypeScript. The existing Python plugins continue to work unchanged for Python users. When Strands Python 2.0 ships with WASM bindings, the TypeScript intervention handlers become available in Python, and the Python-only plugins can be retired.
+
+### The `InterventionHandler` Interface
+
+The interface is **event-driven** — handlers declare which lifecycle events they care about, and the framework only calls them for matching events:
+
+```typescript
+abstract class InterventionHandler {
```
mkmeral: isn't this just hooks with a specific return type? can't we just have hooks support it?

designs/0007-intervention-primitive.md
zastrowm: ~~Purely out of curiosity, did we have any other contenders for the name?~~

~~I can't come up with better than intervention, but it's not necessarily where I would look for something like "permissions" which is where I'd see this going.~~

Appendix E has this

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+    └─ LLM Steering:        (never reached — saved ~100ms)
+```
+
+**Deny** short-circuits immediately. **Guide** accumulates across handlers. **Interrupt** pauses if no handler denied.
+
+No single handler catches everything — the value is in composition. See [Appendix C](#appendix-c-coverage-matrix) for the full matrix.
+
+### Interrupt: Human-in-the-Loop
+
+When a handler returns `Interrupt`, the registry calls `event.interrupt()` — the SDK's native mechanism for pausing execution and requesting human input. The agent pauses, the caller prompts the human, and on resume the handler's `evaluate()` runs again with the human's response available. Any handler can return `Interrupt` for any reason — authorization for consent-gated tools, a guardrail on flagged-but-ambiguous content, steering when it's unsure. See [`DEMO_CONSENT_WALKTHROUGH.md`](./demos/DEMO_CONSENT_WALKTHROUGH.md) for a worked example using Cedar consent policies.
```
JackYPCOnline, line 200: What if handler already ran before human- in -the -loop ? it would re-run?

designs/0007-intervention-primitive.md
lizradway: Steering uses guide/interrupt through hooks but has some workarounds specific to that one plugin to allow the exception to bubble up. Guide cancels the tool and puts feedback into the cancel message to aim for the LLM to retry (hopefully better) whereas Interrupt throws an exception that propagates up since it is outside the try catch boundary. 

If Datadog or a Cedar plugin wants "guide" semantics, they'd have to reverse engineer the same pattern, which to me, begs the need for a common abstraction.

designs/0007-intervention-primitive.md
lizradway: hahahah see the last appendix! lots of options

designs/0007-intervention-primitive.md
mkmeral: how would it be different for `Deny`? do we cancel the event loop, do we stop calling the LLM?

designs/0007-intervention-primitive.md
lizradway: i like that

designs/0007-intervention-primitive.md
mkmeral: i meant more as, can't the plugins use logs/traces to expose this info? Is this all about just making it into a single format?

designs/0007-intervention-primitive.md
lizradway: see final appendix for other options! lots of fun ones

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+```
+
+New event types can be supported without changing the base interface. For the rationale behind this design (vs. fixed methods per hook point), see [Appendix B](#appendix-b-interface-design-rationale).
+
+### The `InterventionRegistry`
+
+The framework provides an `InterventionRegistry` that wires handlers into the Strands hook system. It registers one callback per event type, dispatches to all matching handlers in registration order, and applies conflict resolution:
+
+- **Deny** short-circuits immediately — remaining handlers never run
+- **Guide** accumulates across handlers — feedback from all handlers is concatenated, then the tool is cancelled with the combined guidance so the agent can retry
```
gautamsirdeshmukh: Definitely agree we should feed retry instructions back to the agent - what approach did you have in mind to prevent infinite retries? If the same logically-sound step fails to unblock the task X times, should we immediately stop and let the user know? Should the agent continue reasoning on alternate approaches for X amount of time before surfacing to the user? One way or another seems like we may need to offer some control over this Guide/retry case

designs/0007-intervention-primitive.md
mkmeral: I was thinking guardrails tbh

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+abstract class InterventionHandler {
+    abstract name: string;
+    abstract handles(): Set<typeof HookEvent>;  // e.g. BeforeToolCallEvent, AfterModelCallEvent
+    abstract evaluate(event: HookEvent): Promise<InterventionAction>;
+}
+```
+
+New event types can be supported without changing the base interface. For the rationale behind this design (vs. fixed methods per hook point), see [Appendix B](#appendix-b-interface-design-rationale).
+
+### The `InterventionRegistry`
```
JackYPCOnline, line 150: Does this single registry have a lot responsibilities?

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+
+**3. Unified audit.** When debugging "why was this tool call blocked?", there's no single place to look — each plugin logs on its own (if it logs at all). The intervention registry records every handler's decision — handler name, event type, action, reason, principal, tool — into one stream at the framework level, regardless of whether individual services also log on their side.
+
+**4. Explicit conflict resolution.** If Cedar allows and Galileo denies, what happens? Today: both run, both mutate the event, last write wins. The outcome depends on plugin registration order, which is implicit. The intervention registry makes conflict resolution explicit: Deny wins over everything, Guide accumulates, Interrupt pauses if nothing denied.
```
notowen333: Since the relationships between the interrupts have an inter-tangled behavior, are we confident that the current set is exhaustive? (With the goal of avoiding a breaking change in the future)

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+    // Cedar and steering fire independently — no way to skip steering when Cedar denies
+})
+```
+
+With interventions as a first-class parameter:
+
+```typescript
+const agent = new Agent({
+    tools: [queryDatabase, sendEmail],
+    interventions: [cedar, guardrails, steering],  // cheapest first
```
notowen333, line 116: nit:
`intervention: [steering]` reads rough as a happy path

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+
+### The `InterventionHandler` Interface
+
+The interface is **event-driven** — handlers declare which lifecycle events they care about, and the framework only calls them for matching events:
+
+```typescript
+abstract class InterventionHandler {
+    abstract name: string;
+    abstract handles(): Set<typeof HookEvent>;  // e.g. BeforeToolCallEvent, AfterModelCallEvent
+    abstract evaluate(event: HookEvent): Promise<InterventionAction>;
```
zastrowm: I'm not entirely a fan of returning hooks as "What I support"; this conflates low-level hooks (the implementation) with high-level Interventions.

designs/0007-intervention-primitive.md
```diff
@@ -0,0 +1,454 @@
...
+
+<details>
+<summary><strong>Appendix E: Naming Alternatives</strong></summary>
+
+"Intervention" is the working name, but it carries a connotation of something going wrong (medical intervention, addiction intervention). Authorization isn't corrective — it's a gate. Alternatives worth considering:
+
+| Name | API | Pros | Cons |
+|---|---|---|---|
+| **Intervention** | `Agent(interventions=[...])` | Descriptive — something intervenes in the loop | Implies misbehavior. Unfamiliar as a CS primitive |
+| **Middleware** | `Agent(middleware=[...])` | Instantly familiar to every web engineer. Accurate | May imply single-request linear chain, not multi-event |
```
zastrowm, line 499: I'd avoid Middleware to allow us to use it for something else going forward

designs/0007-intervention-primitive.md
lizradway: For both `Deny` and `Guide`, the tool call is cancelled. In deny, the agent just explains explains denial reason to the user without retry and continues in the event loop. In guide, the agent receives feedback, cancels the current tool, and retries with feedbcak.

designs/0007-intervention-primitive.md
lizradway: They can and some do (Datadog). Each plugin decides independently what to log, how to structure it, and whether to log at all. (Steering doesn’t emit structured decision logs, handrolled auth might log to a file, Galileo logs to its own server, etc)

This is essentially providing a shared schema that is default rather than opt in and can then be exported to OTEL.

lizradway: Updated doc to note/be clear about this specifically

designs/0007-intervention-primitive.md
lizradway: Yeah i don't love the idea of hard coding this check into each plugin

lizradway: For an example, if steering sets `cancelTool` to "Email body is too short, add more context", the agent retries, but stopping on that  `cancelTool` from Steering would mean we wouldn't see if the Cedar plugin vended a "DENY", which stops the user altogether from being able to send emails, which would force another unnecessary loop to detect.

designs/0007-intervention-primitive.md
lizradway: Plugins already do! Intervention audit log is the common schema for these, which can then be exported to OTEl

designs/0007-intervention-primitive.md
lizradway: Followed up in person doc review. We decided to bring this to Python 1.0 & Typescript 1.0 concurrently to allow access asap to customer base. Will update doc accordingly.

designs/0007-intervention-primitive.md
lizradway: Added a section in the appendix to cover this. You can build all of this with vanilla hooks today (and is in our existing plugins), but I would argue this pattern is common enough in our existing plugins (and an important enough agent-primitive) that it justifies a primitive in our SDK.

lizradway: The existing patterns that are continuously reimplemented being: short circuiting, distinguishing deny/guide/etc, and ordering evaluation

designs/0007-intervention-primitive.md
lizradway: yes

designs/0007-intervention-primitive.md
lizradway: Currently, Steering does not have any logic for this.

I'd be open to setting a maxRetries limit on guide until it falls back to interrupt or something of the sort.

designs/0007-intervention-primitive.md
lizradway: It does 3 things:
1. Route events to matching handlers
2. Compose response (ie. Short circuit on deny/ accumulate guide, etc)
3. Record each decision

designs/0007-intervention-primitive.md
lizradway: Interesting thought -- i think per the Bedrock Guardrails example brought up by @zastrowm notes the need for a 5th Intervention Action for replacement/redaction/any code ran on the output automatically. I like `Transform` I think. Any thoughts on that?

designs/0007-intervention-primitive.md
lizradway: strandsSteering?

designs/0007-intervention-primitive.md
lizradway: Yeah that makes sense. Thoughts on dropping `handles` completely to instead opt in to overriding events and continuing with "Proceed" on any non-overridden events?


```
class CedarAuth extends InterventionHandler {
    async beforeToolCall(event): Promise<InterventionAction> {
        return new Deny("not authorized");
    }
    // everything else defaults to Proceed
}
```

designs/0007-intervention-primitive.md
lizradway: yes -- priority goes 
1. Deny: short-circuits immediately, nothing else runs
2. Interrupt: short-circuits, pauses for human input
3. Transform:  applies modification, pipeline continues (next handler sees transformed content)
4. Guide: accumulates across handlers, tool cancelled with combined feedback on retry
5. Proceed: continues to next handler

designs/0007-intervention-primitive.md
zastrowm: I think this makes the better DevX

designs/0007-intervention-primitive.md
```diff
@@ -33,38 +34,37 @@
...
-| **Interrupt** | Pause for human input |
-
-**Deny** is new — steering today only has Proceed/Guide/Interrupt. Authorization needs a hard block that means "you are not allowed, period."
+| Action | Meaning | Details |
+|--------|---------|---------|
+| **Proceed** | Allow | Tool executes or model response is accepted. |
+| **Deny** | Hard block, no retry | New action — steering today only has Proceed/Guide/Interrupt. Authorization needs a hard block that means "you are not allowed, period." |
+| **Guide** | Cancel + feedback for retry | Tool is cancelled, agent receives combined feedback from all handlers and retries with a different approach. |
+| **Interrupt** | Pause for human input | Agent pauses via the SDK's native interrupt system. Human responds, agent resumes. |
+| **Transform** | Modify content and continue | Handler returns modified content (e.g. Bedrock Guardrails redacting PII). Framework applies it, pipeline continues. Later handlers see the transformed content, not the original. |
```
zastrowm, line 67: Are we sure we don't want to re-run the intervention handlers in this case?

designs/0007-intervention-primitive.md
```diff
@@ -135,51 +119,67 @@ const agent = new Agent({
...
 ```typescript
-abstract class InterventionHandler {
-    abstract name: string;
-    abstract handles(): Set<typeof HookEvent>;  // e.g. BeforeToolCallEvent, AfterModelCallEvent
-    abstract evaluate(event: HookEvent): Promise<InterventionAction>;
+class CedarAuth extends InterventionHandler {
+    name = "cedar-auth";
+
+    // Only beforeToolCall is overridden — framework knows this handler only cares about BeforeToolCall
+    async beforeToolCall(event: BeforeToolCallEvent): Promise<InterventionAction> {
```
zastrowm, line 133: +1 - I remember this is what Steering did, so I think this makes sense

zastrowm (approved):

designs/0007-intervention-primitive.md
lizradway: I don't see a use case, but lmk if you have thoughts? Since transforms are additive, redaction/transformations dont change whether the user is authorized, so re-running earlier handlers adds latency/ risks infinite loop for the same answer.

 if you have a guardrail that needs to see the content after PII redaction, put it after the Bedrock Guardrails handler in the interventions array. Handlers run in order, so later handlers see whatever earlier handlers transformed.

PR #780, opened 2026-04-24
docs(designs): background tasks

gautamsirdeshmukh:
## Type of Change

- New content
- Content update/revision
- Typo/formatting fix
- Bug fix
- Other (please describe):

## Checklist

----


designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+### The agent can now drive concurrency on its own
+
+The model can dispatch multiple tools simultaneously, and they all begin executing immediately. As each result arrives, the model can react and adjust its strategy in real time, triggering follow-up work as needed or cancelling tasks that are no longer required. No predefined topology is required – the model's dispatch strategy emerges from its own reasoning.
+
+### A single agent instance can now run concurrent work
+
+The SDK now provides a built-in mechanism to "fork" an agent (create an independent copy) and run it alongside the original. No manual cloning, no lock conflicts, no coordination overhead.
+
+**Zero overhead when not used.** Agents that don't configure `backgroundTools` pay no cost. No system prompt augmentation is injected, no management tools are registered, no token or context overhead is added. The decision points check `taskManager.size` and `_backgroundToolNames` and short-circuit immediately — no forks, no queues, no settlement checks. The agent loop behaves identically to today.
```
JackYPCOnline, line 77: is that mean you define it as tool not aysnc subagent task

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+Two paths in the modified loop converge on the same blocking wait:
+
+1. **Decision Point C**: The model ends its turn (no foreground work remaining) but background tasks are still in flight. Without the wait, the loop would exit and the `finally` block would cancel all pending tasks.
+2. **All-ACK turns**: The model calls multiple tools and *all* of them are background tasks — every tool result is merely an acknowledgement. Without the wait, the model would re-enter the loop with nothing but ACKs to reason about, risking hallucinated results or redundant re-dispatches.
+
+In both cases, the agent blocks until the next background task settles, injects the result into the conversation, and re-enters the loop. See [Cancellation](#cancellation) for the safety bounds that prevent indefinite waits.
+
+### How the Model Sees Background Tasks
+
+Background tools appear in the model's tool definitions identically to foreground tools — same name, same description, same input schema. There is no schema-level async marker. The model learns which tools run asynchronously and how to interact with them solely from the system prompt augmentation described below.
```
mkmeral, line 110: why the system prompt, why not just add it to the tool descriptions?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+## How It Works
+
+### Current Agent Loop
+
+![Current Agent Loop](0009-current-agent-loop.png)
+
+### Modified Agent Loop
+
+![Modified Agent Loop](0009-modified-agent-loop.png)
```
JackYPCOnline, line 87: Can this background tool makes recursive tool call? what will it be for the case single tool is background tool while others are not.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+                            input: { query: "research latest advancements in agentic AI" } }
+[user]       tool_result: { toolUseId: "tu-1", status: "success",
+                            content: "Background task dispatched" }
+```
+
+The ACK pairs with the original `tool_use` via `toolUseId`, just like any foreground tool result. The system prompt tells the model this is not a real result — actual output will arrive later.
+
+#### Result Notification
+
+When the background task completes, its result is injected into the conversation as a user text message — not a `tool_result`, since there is no `tool_use` to pair it with. The `toolUseId` from the original dispatch is echoed for correlation:
```
mkmeral, line 157: if I am having a conversation with the agent though, is this a behavior i want? 

or if the agent is in the middle of an event loop writing code for example

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+![Modified Agent Loop](0009-modified-agent-loop.png)
+
+Legend: <span style="color:#28a745">**green** = new</span>, **gray** = existing (unchanged from current loop)
+
+### Three Decision Points
+
+| Point | Location | Blocking | Purpose |
+|-------|----------|----------|---------|
+| **A** | Start of each loop cycle | No | Pop any background tasks that finished (success, error, or cancelled) since the last cycle and inject their results into the conversation as user messages. Proceeds immediately if none have settled. |
+| **B** | Per tool, during dispatch | No | For each tool the model calls, check if it's designated as a background tool. If yes: fork the agent (or queue if `maxConcurrentBackgroundTasks` is reached), dispatch the tool on the fork, and return an immediate ACK to the model. If no: execute the tool inline as normal. The agent continues without waiting for background results. |
```
pgrayy, line 96: You mention forking the agent but dispatching the tool on the fork. Do we need to fork the entire agent? Also, do we need to use forks at all? What if tools were responsible for returning an ack if they dispatch a background process. I think in most scenarios it is going to be the case that a tool calls an api that is running a process on a separate server and thus the fork wouldn't be necessary.

designs/0009-background-tasks.md
gautamsirdeshmukh: See Appendix G -- tested out updating tool spec description as well as a few other alternatives before settling on system prompt augmentation

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+### The agent no longer blocks on every tool
+
+Tools can now be executed asynchronously and independently, no longer blocking the agent. Results are injected back into the conversation as they finish – the model sees them on its next turn.
+
+### The agent can now drive concurrency on its own
+
+The model can dispatch multiple tools simultaneously, and they all begin executing immediately. As each result arrives, the model can react and adjust its strategy in real time, triggering follow-up work as needed or cancelling tasks that are no longer required. No predefined topology is required – the model's dispatch strategy emerges from its own reasoning.
+
+### A single agent instance can now run concurrent work
```
poshinchen, line 73: Does it work for agent as tools too?

designs/0009-background-tasks.md
opieter-aws, line 77: Can you add `backgroundTools` to the definitions?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+**How it works:**
+- Calling one returns: `Background task dispatched` (started immediately) or `Background task queued` (waiting for a slot).
+- Do not fabricate results. You are not blocked — continue working or yield your turn.
+- Results arrive as user messages:
+
+[Background Task Result]
+tool: <tool_name>
+toolUseId: <tool_use_id>
+status: success|error|cancelled
```
JackYPCOnline, line 129: this can't be interrupted right?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+2. **All-ACK turns**: The model calls multiple tools and *all* of them are background tasks — every tool result is merely an acknowledgement. Without the wait, the model would re-enter the loop with nothing but ACKs to reason about, risking hallucinated results or redundant re-dispatches.
+
+In both cases, the agent blocks until the next background task settles, injects the result into the conversation, and re-enters the loop. See [Cancellation](#cancellation) for the safety bounds that prevent indefinite waits.
+
+### How the Model Sees Background Tasks
+
+Background tools appear in the model's tool definitions identically to foreground tools — same name, same description, same input schema. There is no schema-level async marker. The model learns which tools run asynchronously and how to interact with them solely from the system prompt augmentation described below.
+
+When any background tools are passed to the agent, the SDK auto-generates and appends the following block to the system prompt:
```
opieter-aws, line 112: So background tools need to be configured by the user? Why can't the agent make the decision to run a tool in the background or not?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+---
+
+## Design Decisions & Alternatives
+
+### Why backgroundTools as a First-Class Config (And Not Something Else)
+
+Two alternative approaches to giving the model async dispatch capability:
+
+**1. Meta-tool: `run_in_background(tool_name, args)`**
```
lizradway, line 658: I'm more for this tool wrapper -- it enables the model driven approach that is the backbone of strands.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+## New Primitives
+
+### BackgroundTask
+
+`BackgroundTask` is the foundational unit of background work in the SDK. Every background operation, whether dispatched by the model via `backgroundTools` or by the developer via `invokeBackground()`, is represented as a `BackgroundTask`. It wraps a Promise with synchronous status inspection and implements `PromiseLike` for direct `await`, enabling both the agent loop and the developer to check status, read results, or cancel work without blocking.
+
+```typescript
+type TaskStatus = 'queued' | 'inProgress' | 'success' | 'error' | 'cancelled'
+
+class BackgroundTask implements PromiseLike<unknown> {
```
mkmeral, line 199: how does this correlate to MCP's task definition?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+**Prior art: Mastra's dynamic dispatch.** Mastra solves the "same tool, both modes" problem by allowing the model to include a `_background` field in tool call args to override background/foreground per-call. This is a valid approach that adds flexibility, but it adds a hidden parameter to every tool's input schema, requires the model to learn when to use it, and means the developer can't guarantee a tool will always run in a specific mode. We chose static assignment for v1 because it's simpler, predictable, and lets the developer reason about fork safety at construction time. Dynamic dispatch via an opt-in allowlist (developer pre-approves which tools can be dynamically backgrounded, model decides per-call) is the natural extension path if static assignment proves too restrictive.
+
+**3. Task management tool: `manage_tasks({ action: "create" | "status" | "stop" | "get_result", ... })`**
```
lizradway, line 681: this seems vaguely coupled to https://github.com/strands-agents/tools/pull/389, which is the context management side of self-managed tools (whereas this is the lifecycle side). i would say they are very intertwined though and worth scoping as we look more into meta tools for context

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+| Created but `maxConcurrentTasks` reached — waiting for a slot | `'queued'` |
+| Slot available, fork created, tool executing | `'inProgress'` |
+| `cancel()` called | `'cancelled'` |
+| Promise rejected | `'error'` |
+| Promise resolved, tool returned `status: 'error'` | `'error'` |
+| Promise resolved, tool returned `status: 'success'` | `'success'` |
+
+Task state is tracked via an internal discriminated union rather than promise state, because the agent loop must check settlement without blocking — a raw Promise offers no synchronous status inspection. The union carries the associated data for each state (result value, error, cancellation reason), eliminating the need for separate flags. When `cancel()` is called, status transitions to `'cancelled'` immediately regardless of current state (queued or inProgress). No-op if the task has already settled. See [Cancellation](#cancellation) for the full API.
+
+#### TaskManager
```
opieter-aws, line 252: Nomenclature nit: should this be `BackgroundTaskManager` if it manages `BackgroundTasks`?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+The model can dispatch multiple tools simultaneously, and they all begin executing immediately. As each result arrives, the model can react and adjust its strategy in real time, triggering follow-up work as needed or cancelling tasks that are no longer required. No predefined topology is required – the model's dispatch strategy emerges from its own reasoning.
+
+### A single agent instance can now run concurrent work
+
+The SDK now provides a built-in mechanism to "fork" an agent (create an independent copy) and run it alongside the original. No manual cloning, no lock conflicts, no coordination overhead.
+
+**Zero overhead when not used.** Agents that don't configure `backgroundTools` pay no cost. No system prompt augmentation is injected, no management tools are registered, no token or context overhead is added. The decision points check `taskManager.size` and `_backgroundToolNames` and short-circuit immediately — no forks, no queues, no settlement checks. The agent loop behaves identically to today.
+
+## How It Works
```
mkmeral, line 79: For this to work, do we need to keep the process/runtime open?

designs/0009-background-tasks.md
JackYPCOnline, line 157: will this bring side effects across different models?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+A NASA mission controller dispatches probes to 3 planets and researches each one. 6 tool calls, 1 wave.
+
+#### Cross-Model Summary
+
+| Model | Runs | Avg Standard | Avg Background | Avg Speedup | σ | Token Δ | Validations |
+|-------|------|-------------|----------------|-------------|---|---------|-------------|
+| Sonnet 4.6 | 5 | 86.5s | 31.4s | **2.80x** | ±0.40 | 4.4% | 110/110 |
+| Haiku 4.5 | 5 | 79.6s | 28.2s | **2.89x** | ±0.40 | 18.2% | 110/110 |
+
+Context growth (avg messages): 6-7 standard vs 8-10 background. The additional messages are injected background task results — expected behavior, not overhead. Input tokens increase 11-40% due to the model seeing injected results across multiple turns rather than in a single batch. See [Context Management](#context-management).
```
lizradway, line 927: a 40% increase in tokens is highly noteworthy. that is a huge cost

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+```
+
+#### backgroundTools Config
+
+`backgroundTools` accepts the same types as `tools` — `Tool`, `McpClient`, `Agent`, `Graph`, `Swarm`, or nested arrays. Anything that can be a tool can be a background tool.
+
+```typescript
+const agent = new Agent({
+  tools: [calculateMetrics, formatReport],
+  backgroundTools: [searchWeb, analyzeData, researcher],
```
opieter-aws, line 321: Is there a different way we can define this interface that makes it clear that `tools` is the superset of `tools` and `backgroundTools`?

designs/0009-background-tasks.md
mkmeral: alternatively, can we do a strategy where we update the tool description, and say `There are 2 async task results available for ...`, and let it be a bit more model driven instead of default notification? 

how do the possibly different contexts interact with each other?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+              tool: searchWeb
+              toolUseId: tu-1
+              status: success
+
+              <tool output>"
+```
+
+This is the structural asymmetry at the heart of background tasks: the ACK uses native `tool_result` pairing, but the real result arrives as a plain text message. This is why the system prompt augmentation is necessary — see [Appendix G](#appendix-g-system-prompt-augmentation-rationale).
+
+#### Model-Driven Task Management
```
pgrayy, line 170: I worry this could lead to more token usage. Does the model need to be the one to poll background tasks? Is that something we can do in the agent loop? Or is it something users could do.

designs/0009-background-tasks.md
gautamsirdeshmukh: Ah, good callout here - this particular figure is from earlier benchmarks, more recent testing after tightening up various aspects of the mechanism showed far less bloat. Will update this

designs/0009-background-tasks.md
lizradway: we are augmenting tool results for https://github.com/strands-agents/sdk-python/pull/2162 so i think this does have precedent in the sdk.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+The model receives an ACK for every dispatched tool immediately regardless of queue state. The distinction between "dispatched" and "queued" ACKs lets the model know whether work has started.
+
+Key methods:
+
+- **`enqueue()`** — the entry point for background dispatch. Creates a `BackgroundTask`, either starts it immediately or queues it based on concurrency, and returns the task handle with the appropriate ACK.
+- **`popCompleted()`** — returns and removes all settled tasks from the registry. Triggers queue drain if slots opened. This is the only way tasks leave the registry, ensuring no task is accidentally lost or processed twice.
+- **`cancel(id)`** — cancel a specific task by its internal ID. Works on both queued and running tasks. Cancelling a queued task removes it from the queue without ever forking.
+- **`cancelByToolUseId(toolUseId)`** — cancel a specific task by its `toolUseId`. This is the primary path for model-driven cancellation, since the model knows `toolUseId` from its own `tool_use` blocks in conversation history.
```
mkmeral, line 284: should this be a task manager method though? i think it should be part of the tool, and it should just translate to `cancel(id)`

designs/0009-background-tasks.md
lizradway: plus one -- autonomy feels like the main/only gain of background tasks to me, especially given that is was significantly slower than graph in the examples below

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+|---|---|
+| **Conversation** (messages, app state) | Deep-copied from parent. Independent after forking. Background tool forks start empty ([Context Management](#context-management)). |
+| **Model and tools** | Shared reference. Same model, tool registry, and MCP clients. |
+| **System prompt** | Copied from parent. Overridable via `ForkOptions`. |
+| **Hooks** | Rebuilt with propagating callbacks only ([Hook Propagation](#hook-propagation)). |
+| **Session** | Dropped. Forks do not persist to the parent's session. |
+| **Execution** (invocation lock, cancellation, metrics) | Fresh. Each fork can be invoked and cancelled independently. |
+| **Task management** | Fresh `TaskManager`, same config. Fork manages its own background tasks. |
+
+Messages are deep-copied by default. For background tool forks specifically, the SDK automatically passes `messages: []` at Decision Point B — see [Context Management](#context-management) for why and how to opt out via `inheritMessages`.
```
opieter-aws, line 368: This confuses me. They are copied by default, but not by default for background tool forks? Can you clarify here what the difference scenarios are?

designs/0009-background-tasks.md
lizradway: plus one, if the same tool is included in tools and background tools, what does that mean??

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+agent.addHook(BackgroundTaskResultEvent, (event) => {
+  if (event.error) {
+    console.log(`Task ${event.taskName} failed in ${event.durationMs}ms: ${event.error.message}`)
+  }
+})
+```
+
+#### backgroundTools Config
+
+`backgroundTools` accepts the same types as `tools` — `Tool`, `McpClient`, `Agent`, `Graph`, `Swarm`, or nested arrays. Anything that can be a tool can be a background tool.
```
mkmeral, line 316: isn't this more of a tool concern though? i'd expect more like `@tool(background=True)` or something 🤔

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+| **Fork** | An independent copy of the agent created via `fork()`. Has its own conversation, execution lock, and task manager, but shares the parent's model client and tool registry. The isolation primitive that makes concurrent execution safe. |
+| **Decision Point** | One of three locations in the modified agent loop where background task logic is injected. **A** (top of cycle): pop settled results. **B** (per tool): fork and dispatch or execute inline. **C** (end of turn): wait if tasks are pending. |
+| **ACK** | The immediate `tool_result` returned to the model when a background tool is dispatched. Contains "Background task dispatched" or "Background task queued." Not a real result — the actual output arrives later via injection. |
+| **Injection** | The mechanism by which background task results enter the parent's conversation. Results are appended as user text messages with a `[Background Task Result]` prefix and `toolUseId` for correlation. |
```
poshinchen, line 29: Will the model hallucinate the "late" tool results?

designs/0009-background-tasks.md
pgrayy: I'm thinking of the following scenario:
1. Model emits a tool use.
2. Tool runs in background.
3. No other tool is executing and so no other work is being done in the loop.
4. We exit out of the agent loop with a handler so the user can decide when to reinvoke.
     1. One advantage here is that if no work is being done locally, the user can save on compute and shut down their entire process.
5. Once ready, the user reinvokes the agent.
6. Agent returns back to the tool call to retrieve the result.
7. The result is sent to the model.
8. The model isn't aware that the tool was executed in the background.

Your approach however allows for the model to continue processing concurrent results and follow up on background results. I can see then why passing those as a plain user message rather than a tool result could be helpful. Still curious though if there is a way to still enable passing back as a tool result.

designs/0009-background-tasks.md
lizradway: i like this alot better as well

lizradway: ideally we'd default background to false and allow agents to run tasks in background unless explicitly configured otherwise, i'd imagine?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+  tools: [calculateMetrics, formatReport],
+  backgroundTools: [searchWeb, analyzeData, researcher],
+})
+```
+
+The model sees background tools as normal tools in its tool definitions. A system prompt augmentation explains the async contract — see [How the Model Sees Background Tasks](#how-the-model-sees-background-tasks) for the full prompt block and message format details.
+
+#### fork()
+
+`fork()` creates an independent copy of the agent that can be invoked concurrently with the original. It is both the isolation primitive that makes background tasks possible (each dispatch at Decision Point B creates a fork) and a standalone capability for developers who want to parallelize work using the same agent configuration.
```
JackYPCOnline, line 329: Might need more justification here?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+Each fork has its own `AbortController`. Cancellation flows in one direction:
+
+- **Parent exit → forks cancelled.** When the parent's agent loop exits (normal completion, error, or cancel), its `finally` block calls `taskManager.cancelAll()`, which fires each background fork's abort signal. No orphaned forks.
+- **Fork cancelled → parent unaffected.** A fork's abort controller is independent. Cancelling a fork (via `BackgroundTask.cancel()` or TTL) does not propagate to the parent.
+- **Manual forks.** For forks created directly by the developer (not via `backgroundTools` or `invokeBackground`), the developer is responsible for cancellation — there is no automatic cleanup.
+
+##### Fork depth guard
+
+A configurable depth limit (default: 20, set via `maxForkDepth` on `AgentConfig`) prevents infinite recursive forking — for example, a background tool that itself dispatches background tools. Throws if exceeded.
```
opieter-aws, line 382: Is 20 not a bit excessive here? With 10 forks, all 20 forks deep, that's 200 agents, if the fork is only allowed to spin up 1 fork (can't find the default here). How do we keep cost managed?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+| Promise resolved, tool returned `status: 'success'` | `'success'` |
+
+Task state is tracked via an internal discriminated union rather than promise state, because the agent loop must check settlement without blocking — a raw Promise offers no synchronous status inspection. The union carries the associated data for each state (result value, error, cancellation reason), eliminating the need for separate flags. When `cancel()` is called, status transitions to `'cancelled'` immediately regardless of current state (queued or inProgress). No-op if the task has already settled. See [Cancellation](#cancellation) for the full API.
+
+#### TaskManager
+
+`TaskManager` is the lifecycle manager for `BackgroundTask` instances created during the `backgroundTools` dispatch path ([Decision Point B](#three-decision-points)). It owns settlement detection, cancellation, and cleanup. Each agent instance holds its own `TaskManager`; forks get fresh instances with the same config. This isolation ensures a fork's background tasks are the fork's responsibility — the parent only sees the fork itself as one task, never the sub-tasks the fork may spawn internally.
```
pgrayy, line 254: Do we have to manage our own task manager? This seems heavy and task management has already been solved.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+- **`popCompleted()`** — returns and removes all settled tasks from the registry. Triggers queue drain if slots opened. This is the only way tasks leave the registry, ensuring no task is accidentally lost or processed twice.
+- **`cancel(id)`** — cancel a specific task by its internal ID. Works on both queued and running tasks. Cancelling a queued task removes it from the queue without ever forking.
+- **`cancelByToolUseId(toolUseId)`** — cancel a specific task by its `toolUseId`. This is the primary path for model-driven cancellation, since the model knows `toolUseId` from its own `tool_use` blocks in conversation history.
+- **`cancelByName(name)`** — cancel all in-progress and queued tasks with a given tool name. Returns the count. Available to both the model (via `cancel_background_task`) and developers directly.
+- **`cancelAll()`** — cancel everything (running and queued) and clear the registry. Called by the agent loop's `finally` block on exit.
+
+#### Events
+
+Three new hookable events:
```
mkmeral, line 290: who hooks into these? what are the use cases?

designs/0009-background-tasks.md
opieter-aws: How do we ensure multiple forks aren't converging to all doing the same task accidentally?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+  Third injection:
+  [user]       text: "[Background Task Result]
+                      tool: searchWeb
+                      toolUseId: tu-1
+                      status: success
+                      ..."
+```
+
+The model must not assume results arrive in dispatch order. `toolUseId` correlates each result to its original `tool_use` block.
```
poshinchen, line 1496: We will implement this logic right? instead of letting the model find and map by itself.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+})
+```
+
+The model sees background tools as normal tools in its tool definitions. A system prompt augmentation explains the async contract — see [How the Model Sees Background Tasks](#how-the-model-sees-background-tasks) for the full prompt block and message format details.
+
+#### fork()
+
+`fork()` creates an independent copy of the agent that can be invoked concurrently with the original. It is both the isolation primitive that makes background tasks possible (each dispatch at Decision Point B creates a fork) and a standalone capability for developers who want to parallelize work using the same agent configuration.
+
+Why it's needed: `invoke()` on the same agent instance acquires an `_isInvoking` lock. A second concurrent call throws `ConcurrentInvocationError`. This is a deliberate safety rail — concurrent writes to the same `messages` array would corrupt conversation state. `fork()` gives each concurrent invocation its own messages, state, and lock:
```
mkmeral, line 331: why are we invoking the same agent though? why not just use `use_agent` like sub-agent?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+- **Standard:** Single coordinator, all 6 sub-agents as `tools`. Sequential.
+- **Background:** Single coordinator, all 6 sub-agents as `backgroundTools`. Model discovers parallelism.
+- **Graph:** Explicit DAG with 6 nodes and 6 edges. Graph engine handles parallelism.
+
+| Approach | Wall Clock | vs Standard | Output Length |
+|----------|-----------|-------------|---------------|
+| Standard | 98.1s | baseline | 4,006 chars |
+| Background | 66.8s | **1.47x faster** | 3,371 chars |
+| Graph | 34.8s | **2.82x faster** | 3,866 chars |
```
lizradway, line 1035: graph is alot faster, i feel like swarm would be as well and would address the same usecase. why could we not add async support to swarm?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+try {
+  await task
+} catch (e) {
+  if (e instanceof TaskCancelledError) {
+    // Intentional cancellation, not a failure
+  }
+}
+```
+
+Cancellation follows the same pattern as `fetch` with an aborted `AbortSignal` — intentional cancellation is a rejection, not a silent resolve. The synchronous getters (`status`, `result`, `error`) enable inspection without awaiting.
```
notowen333, line 235: This means that we need to thread through the signal into the tool. If possible I'd avoid handling more than one abort signal in addition to the existing top level agent signal

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+#### TaskManager
+
+`TaskManager` is the lifecycle manager for `BackgroundTask` instances created during the `backgroundTools` dispatch path ([Decision Point B](#three-decision-points)). It owns settlement detection, cancellation, and cleanup. Each agent instance holds its own `TaskManager`; forks get fresh instances with the same config. This isolation ensures a fork's background tasks are the fork's responsibility — the parent only sees the fork itself as one task, never the sub-tasks the fork may spawn internally.
+
+These settings tune the agent loop's waiting behavior at [the wait state](#the-wait-state--why-block):
+
+```typescript
+interface TaskManagerConfig {
+  heartbeatMs?: number           // How often to emit BackgroundTaskPendingEvent while waiting (default: 5000ms)
```
notowen333, line 260: maybe `timeBetweenPendingEventMs` or something else closer to the function

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+
+**Zero overhead when not used.** Agents that don't configure `backgroundTools` pay no cost. No system prompt augmentation is injected, no management tools are registered, no token or context overhead is added. The decision points check `taskManager.size` and `_backgroundToolNames` and short-circuit immediately — no forks, no queues, no settlement checks. The agent loop behaves identically to today.
+
+## How It Works
+
+### Current Agent Loop
+
+![Current Agent Loop](0009-current-agent-loop.png)
+
+### Modified Agent Loop
```
notowen333, line 85: High level from the review meeting: I would be interested to see if we could first develop some constructs around async non-blocking tools/work/agents without deeply integrating the feature into the core event loop.

Proving out this async lifecycle management (very likely re-using similar approaches presented in the doc) in a more additive/compatible way could help to iron out the topic and prove out use cases.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,1590 @@
...
+#### The Wait State — Why Block?
+
+Two paths in the modified loop converge on the same blocking wait:
+
+1. **Decision Point C**: The model ends its turn (no foreground work remaining) but background tasks are still in flight. Without the wait, the loop would exit and the `finally` block would cancel all pending tasks.
+2. **All-ACK turns**: The model calls multiple tools and *all* of them are background tasks — every tool result is merely an acknowledgement. Without the wait, the model would re-enter the loop with nothing but ACKs to reason about, risking hallucinated results or redundant re-dispatches.
+
+In both cases, the agent blocks until the next background task settles, injects the result into the conversation, and re-enters the loop. See [Cancellation](#cancellation) for the safety bounds that prevent indefinite waits.
+
+### How the Model Sees Background Tasks
```
pgrayy, line 108: One quick call out, if we return a tool result later to the model, it may not know what to do with it because context could be lost from other messages added while waiting. Or it could be the case the conversation history was summarized while waiting.

PR #783, opened 2026-04-27
docs(designs): background tasks

gautamsirdeshmukh:
## Type of Change

- New content
- Content update/revision
- Typo/formatting fix
- Bug fix
- Other (please describe):

## Checklist

----

### DIFF
--- a/designs/0009-background-tasks.md
+++ b/designs/0009-background-tasks.md
@@ -1,0 +1,301 @@
+# Background Tasks: Async Dispatch for Strands Agents
+
+**Status:** Proposed
+**Date:** 2026-04-27
+**Author:** Gautam Sirdeshmukh
+**Scope:** TypeScript SDK. Python SDK parity to follow.
+
+---
+
+**Contents:** [Problem](#problem) · [Proposal](#proposal) · [Unlocks](#unlocks) · [Design Alternatives](#design-alternatives) · [Extension to Agent Harness](#extension-to-agent-harness) · [Appendix A: Conversation Trace](#appendix-a-conversation-trace) · [Appendix B: Prior Art](#appendix-b-prior-art) · [Appendix C: Landscape](#appendix-c-landscape) · [Appendix D: Background Tasks vs Graph vs Swarm vs CTE](#appendix-d-background-tasks-vs-graph-vs-swarm-vs-cte) · [Appendix E: Toward Interactive Background Agents](#appendix-e-toward-interactive-background-agents)
+
+---
+
+## Problem
+
+Strands is built around the core concept of giving the agent capabilities but letting the model decide how to use them. That philosophy, however, does not extend to execution. The agent loop is synchronous: each cycle, the model calls tools, then waits, unable to reason or make follow-up calls until every result in that round returns.
+
+*The model can decide what to do, but not when. New work always waits for old work to finish, even when the two are unrelated.*
+
+### The agent is blocked while tools are running
+
+A 30-second API call means the agent idles for 30 seconds. A 15-minute MCP tool running a deep research pipeline blocks the agent for the full 15 minutes — whether or not the result is needed before moving on. The agent has no way to make progress on anything else while it waits.
+
+### The model has no control over concurrency
+
+Every concurrency option in Strands requires developer configuration. If the model determines mid-reasoning that two tasks are independent, it has no way to dispatch them and keep reasoning while they run. ConcurrentToolExecutor helps — it drops latency from sum(tools) to max(tools) — but the developer decides what runs concurrently, not the model, and the agent still blocks until the batch completes.
+
+### Unrelated tasks block each other
+
+If task A finishes in 2 seconds and task B takes 30, the model cannot act on A's result until B also completes — even if A and B have nothing to do with each other. A completed result should be actionable the moment it arrives.
+
+---
+
+## Proposal
+
+Background Tasks give the model control over *when* work runs, not just *what*. A single `background()` tool lets the model dispatch work, keep reasoning, and react to results as they arrive:
+
+| Action | Example | Description |
+|--------|---------|-------------|
+| dispatch | `background({ action: "dispatch", tool: "search_web", args: { query: "..." } })` | Dispatch work, receive a task ID immediately |
+| get | `background({ action: "get", taskId: "..." })` | Check task status (polling is never required — results are injected automatically) |
+| cancel | `background({ action: "cancel", taskId: "..." })` | Cancel a running task |
+
+Anything the model can already call can be dispatched as a background task:
+
+| Target | Example |
+|--------|---------|
+| Tool | `background({ action: "dispatch", tool: "search_web", args: { query: "..." } })` |
+| Agent-as-tool | `background({ ..., tool: "research_agent", args: { input: "..." } })` |
+| MCP tool | `background({ ..., tool: "mcp_tool_name", args: { ... } })` |
+| Graph / Swarm | `background({ ..., tool: "pipeline", args: { input: "..." } })` |
+| Remote agent (A2A) | `background({ ..., tool: "remote_agent", args: { input: "..." } })` |
+| Dynamic agent (use_agent) | `background({ ..., tool: "use_agent", args: { prompt: "...", tools: [...] } })` |
+| Any callable | `background({ ..., tool: "x", args: { ... } })` |
+
+The model keeps reasoning and dispatching more work as needed. As results come in, they are delivered as tagged messages appended to the conversation, not as `tool_result` blocks. The dispatch returns a task ID as a normal `tool_result` (satisfying the provider API's synchronous pairing requirement), and the actual result arrives later as a separate message tagged with the originating tool and task ID. The model sees them at the start of its next turn. The tool definition instructs the model to continue working without fabricating results for dispatched tasks.
+
+Background results are subject to the same context window limits and compaction behavior as any other conversation content. Proactive strategies — structured output on sub-agents to reduce result size, batching closely-completing results into a single injection, or queuing results when the context is near capacity — can mitigate this further.
+
+This is the entire model-facing interface. No changes to the agent loop. No system prompt augmentation required. See Design Alternatives for [result delivery](#result-delivery-strategy) and [tool shape](#separate-tools-instead-of-one) alternatives.
+
+---
+
+## Unlocks
+
+### The model is never idle
+
+Work runs in the background while the model keeps reasoning. A coordinator dispatches 4 researcher agents and immediately moves on to structuring its report outline; when results arrive, they slot into a framework the model has already prepared.
+
+### The model drives its own concurrency
+
+The model can parallelize work at runtime without the need for a ConcurrentToolExecutor, Graph, or any developer-configured infrastructure. Given "research these 5 topics," it dispatches all 5 at once and they begin executing concurrently.
+
+### Results are actionable the moment they arrive
+
+Each completed task flows back to the model independently — a fast result is no longer held up by a slow one. The model can incorporate early results, cancel tasks that are no longer needed, and dispatch follow-up work without waiting for the full batch.
+
+### Anything callable can become an async background process
+
+Tools, agents, MCP tools, Graph pipelines, Swarms, remote A2A agents... if the model can call it, it can background it. No special wrappers, no separate APIs. One mechanism for everything.
+
+---
+
+## Design Alternatives
+
+### Config-based (`backgroundTools`)
+
+The developer declares which tools run in the background at construction time. The model calls tools normally — the SDK intercepts calls to background-listed tools and handles them asynchronously:
+
+```typescript
+// Developer configuration
+const agent = new Agent({
+  tools: [quickLookup],
+  backgroundTools: [searchWeb, analyzeData],
+})
+
+// Model calls searchWeb as usual — the SDK backgrounds it automatically
+```
+
+| Pros | Cons |
+|------|------|
+| Developer controls which tools are safe to background (stateful tools, side effects) | Developer decides what's backgroundable, not the model — no per-conversation flexibility |
+| No additional tool in the model's schema | A tool is always background or always foreground — can't adapt at runtime |
+| Model can't misuse it — only pre-approved tools run in the background | Two tool lists to maintain; overlap between them creates ambiguity |
+| Simpler model behavior — fewer failure modes (no hallucinated results, no polling loops) | Requires system prompt augmentation so the model understands async results |
+| | Conflicts with the core Strands principle of letting the model drive decisions |
+
+### Per-tool config
+
+Instead of a meta-tool or a separate list, the background property lives on each tool definition. The model still calls tools normally — the SDK checks the per-tool config and backgrounds the call automatically.
+
+```typescript
+const agent = new Agent({
+  tools: [
+    quickLookup,
+    { tool: searchWeb, background: true },
+    { tool: analyzeData, background: true },
+  ],
+})
+```
+
+| Pros | Cons |
+|------|------|
+| Single source of truth — each tool's config in one place, no duplicate lists | Introduces a wrapper object into the tools array that every consumer must handle |
+| Extensible — the wrapper could carry other per-tool options (TTL, priority) | Developer decides, model can't adapt at runtime |
+| No overlap ambiguity — a tool can't appear in both lists | Changes the shape of the tools array, which is currently a flat list across the SDK |
+| Could default to `false` and let developers opt in per-tool | The model still has no say in what gets backgrounded |
+
+### Separate tools instead of one
+
+Instead of a single `background()` tool with an `action` parameter, the interface could be split into three dedicated tools: `background()` for dispatch, `getTask()` for status, and `cancelTask()` for cancellation.
+
+| Pros | Cons |
+|------|------|
+| Each tool has a focused schema — simpler for the model per-call | Three tools in the registry instead of one |
+| Matches the MCP Tasks protocol shape (`tasks/get`, `tasks/cancel`) | Model must discover and learn three tools instead of one |
+| No `action` parameter to misuse | More surface area to maintain |
+
+Either shape works. We propose the single-tool approach because it keeps the registry minimal and the model only needs to learn one tool name.
+
+### Result delivery strategy
+
+The proposal delivers results as tagged messages in the conversation. Other frameworks take different approaches:
+
+| Strategy | Pros | Cons |
+|----------|------|------|
+| Tagged message injection (this proposal) | No history mutation; simple to implement; model sees results at clean turn boundaries | Model must learn from tool definition that results arrive as messages, not `tool_result` blocks |
+| Retroactive `tool_result` (Mastra) | Natural for the model — standard tool pairing, no new patterns to learn | Mutates conversation history after the model has already reasoned past that point; if compaction removes the original `tool_use`, the retroactive result has nothing to pair with; provider validation risks |
+| Explicit polling (LangChain) | Zero hallucination risk — model actively retrieves results | Token overhead from repeated status checks; wasted model calls when results aren't ready |
+| Status + auto-notification (Claude Agent SDK) | Clear "async_launched" signal reduces hallucination; model can poll or wait | More complex — requires both notification infrastructure and polling tools |
+
+### Model-unaware async
+
+The following is not an alternative to Background Tasks but a complementary approach for a different use case.
+
+The model calls a tool normally. Behind the scenes, the framework intercepts the result, releases the agent process, and waits for the external work to complete. The agent is re-invoked when ready, and the result is injected as if the tool ran synchronously. The model never knows anything was async.
+
+This is the approach taken by the [Strands Durability Plugin (SARK)](https://sark-docs.beta.harmony.a2z.com/sdo-strands-durability-plugin/developer-guide/async-execution). SARK's interrupt-based session release is more applicable to the durable task phase than to the ephemeral mechanism proposed here.
+
+| Pros | Cons |
+|------|------|
+| Zero model complexity — no new tools, no async semantics to learn | Model cannot dispatch multiple tasks concurrently |
+| Releases compute during long waits — no idle GPU/container | Model cannot reason about partial results as they arrive |
+| No risk of hallucinated results or polling loops | Only suited for single long-running tasks, not parallel workloads |
+| Works with any existing tool without modification | Requires platform-level re-invocation infrastructure |
+
+**This approach is complementary, not competing.** `background()` keeps the model productive while multiple tasks run. Model-unaware async releases compute when the model has nothing else to do. In a managed deployment, both compose: the model dispatches background tasks, and the platform releases compute while waiting for results.
+
+---
+
+## Extension to Agent Harness
+
+Background Tasks as proposed are ephemeral: in-process, tracked in memory, scoped to the agent's lifetime. If the agent process dies, the tasks die with it. This is acceptable for interactive sessions and short-lived agentic workflows, but does not scale for long-running production tasks. 
+
+The path from ephemeral to *durable* requires three things: isolated execution (so tasks outlive the agent process), persistent state (so the agent can resume), and an external task store (so task metadata survives). This is what an agent harness provides.
+
+### Ephemeral background tasks
+
+The base-level mechanism. Tracked in memory, scoped to the agent's lifetime, no infrastructure required. Ships as part of this proposal.
+
+### Durable background tasks
+
+The model calls the same `background()` tool. What changes is the infrastructure underneath. Tasks run outside the agent process and survive process death. The agent can resume from where it left off, and task metadata persists across sessions.
+
+Durable background tasks enable fire-and-forget: the agent can dispatch work, exit, and pick up results on its next invocation. They also open the door to ambient, long-running processes — background agents that persist across sessions, react to events, and operate alongside the main agent over extended timeframes.
+
+Two efforts already in progress provide the foundation: the Sandbox abstraction (PR #681) provides isolated execution environments, and the checkpointing mechanism (PR #2190) enables agent state persistence and resumption.
+
+The durable interface maps directly to the MCP Tasks protocol:
+
+| MCP Tasks | Durable background tasks |
+|-----------|--------------------------|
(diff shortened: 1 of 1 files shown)


designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+---
+
+## Proposal
+
+Background Tasks give the model control over *when* work runs, not just *what*. A single `background()` tool lets the model dispatch work, keep reasoning, and react to results as they arrive:
+
+| Action | Example | Description |
+|--------|---------|-------------|
+| dispatch | `background({ action: "dispatch", tool: "search_web", args: { query: "..." } })` | Dispatch work, receive a task ID immediately |
+| get | `background({ action: "get", taskId: "..." })` | Check task status (polling is never required — results are injected automatically) |
```
zastrowm, line 41: Doc feedback: It's a little unclear what "results are injected automatically" at this point; it raises more questions

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+| Tool | `background({ action: "dispatch", tool: "search_web", args: { query: "..." } })` |
+| Agent-as-tool | `background({ ..., tool: "research_agent", args: { input: "..." } })` |
+| MCP tool | `background({ ..., tool: "mcp_tool_name", args: { ... } })` |
+| Graph / Swarm | `background({ ..., tool: "pipeline", args: { input: "..." } })` |
+| Remote agent (A2A) | `background({ ..., tool: "remote_agent", args: { input: "..." } })` |
+| Dynamic agent (use_agent) | `background({ ..., tool: "use_agent", args: { prompt: "...", tools: [...] } })` |
+| Any callable | `background({ ..., tool: "x", args: { ... } })` |
+
+The model keeps reasoning and dispatching more work as needed. As results come in, they are delivered as tagged messages appended to the conversation, not as `tool_result` blocks. The dispatch returns a task ID as a normal `tool_result` (satisfying the provider API's synchronous pairing requirement), and the actual result arrives later as a separate message tagged with the originating tool and task ID. The model sees them at the start of its next turn. The tool definition instructs the model to continue working without fabricating results for dispatched tasks.
```
opieter-aws, line 56: What will the implication of this be on context management features like tool result offloading?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Separate tools instead of one
+
+Instead of a single `background()` tool with an `action` parameter, the interface could be split into three dedicated tools: `background()` for dispatch, `getTask()` for status, and `cancelTask()` for cancellation.
+
+| Pros | Cons |
+|------|------|
+| Each tool has a focused schema — simpler for the model per-call | Three tools in the registry instead of one |
+| Matches the MCP Tasks protocol shape (`tasks/get`, `tasks/cancel`) | Model must discover and learn three tools instead of one |
+| No `action` parameter to misuse | More surface area to maintain |
```
zastrowm, line 137: More chance of conflicts too.

It might be good as a team to align on what we prefer/generally do; that also helps us shape future tools

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+Either shape works. We propose the single-tool approach because it keeps the registry minimal and the model only needs to learn one tool name.
+
+### Result delivery strategy
+
+The proposal delivers results as tagged messages in the conversation. Other frameworks take different approaches:
+
+| Strategy | Pros | Cons |
+|----------|------|------|
+| Tagged message injection (this proposal) | No history mutation; simple to implement; model sees results at clean turn boundaries | Model must learn from tool definition that results arrive as messages, not `tool_result` blocks |
```
zastrowm, line 147: Doc feedback: It would be good to detail what it is before going into the pros & cons

designs/0009-background-tasks.md
mkmeral, line 56: > The model sees them at the start of its next turn

need a bit more explanation here

designs/0009-background-tasks.md
zastrowm, line 147: Any concern about prompt injection?  E.g. we've had model reject data because it seems like prompt injection - do we have any idea if that would/might happen here?

designs/0009-background-tasks.md
pgrayy, line 56: What if we waited to append the tool use until we had the tool result. This is actually what we do in TS already. Then you don't need to send a "task id" tool result and the actual tool result can be packed into a tool result message.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+Each completed task flows back to the model independently — a fast result is no longer held up by a slow one. The model can incorporate early results, cancel tasks that are no longer needed, and dispatch follow-up work without waiting for the full batch.
+
+### Anything callable can become an async background process
+
+Tools, agents, MCP tools, Graph pipelines, Swarms, remote A2A agents... if the model can call it, it can background it. No special wrappers, no separate APIs. One mechanism for everything.
+
+---
+
+## Design Alternatives
```
mkmeral, line 84: can we extend `AgentTool` so tools expose `tool.backgroundable` or sth? so it's not a wrapper but part of the tool, that it can be turned to background? 🤔

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Result delivery strategy
+
+The proposal delivers results as tagged messages in the conversation. Other frameworks take different approaches:
+
+| Strategy | Pros | Cons |
+|----------|------|------|
+| Tagged message injection (this proposal) | No history mutation; simple to implement; model sees results at clean turn boundaries | Model must learn from tool definition that results arrive as messages, not `tool_result` blocks |
+| Retroactive `tool_result` (Mastra) | Natural for the model — standard tool pairing, no new patterns to learn | Mutates conversation history after the model has already reasoned past that point; if compaction removes the original `tool_use`, the retroactive result has nothing to pair with; provider validation risks |
+| Explicit polling (LangChain) | Zero hallucination risk — model actively retrieves results | Token overhead from repeated status checks; wasted model calls when results aren't ready |
```
zastrowm, line 149: Will we support `Explicit polling` as option regardless or no?

zastrowm: > Token overhead from repeated status checks; wasted model calls when results aren't ready

There's a world where we could auto-trim old messages, but I do agree this can be a waste

designs/0009-background-tasks.md
lizradway: plus one

designs/0009-background-tasks.md
zastrowm: > Will we support Explicit polling as option regardless or no?

Oh, I see that Status + auto-notification is what I was suggesting?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+This is the entire model-facing interface. No changes to the agent loop. No system prompt augmentation required. See Design Alternatives for [result delivery](#result-delivery-strategy) and [tool shape](#separate-tools-instead-of-one) alternatives.
+
+---
+
+## Unlocks
+
+### The model is never idle
+
+Work runs in the background while the model keeps reasoning. A coordinator dispatches 4 researcher agents and immediately moves on to structuring its report outline; when results arrive, they slot into a framework the model has already prepared.
```
pgrayy, line 68: How do we handle tool event streaming? If the tool is running in the background, are we still able to stream back intermediate events? Is that something we want.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+### Result delivery strategy
+
+The proposal delivers results as tagged messages in the conversation. Other frameworks take different approaches:
+
+| Strategy | Pros | Cons |
+|----------|------|------|
+| Tagged message injection (this proposal) | No history mutation; simple to implement; model sees results at clean turn boundaries | Model must learn from tool definition that results arrive as messages, not `tool_result` blocks |
+| Retroactive `tool_result` (Mastra) | Natural for the model — standard tool pairing, no new patterns to learn | Mutates conversation history after the model has already reasoned past that point; if compaction removes the original `tool_use`, the retroactive result has nothing to pair with; provider validation risks |
+| Explicit polling (LangChain) | Zero hallucination risk — model actively retrieves results | Token overhead from repeated status checks; wasted model calls when results aren't ready |
+| Status + auto-notification (Claude Agent SDK) | Clear "async_launched" signal reduces hallucination; model can poll or wait | More complex — requires both notification infrastructure and polling tools |
```
mkmeral, line 150: Can you give a bit more context here? how do they handle notification? Why didn't we go with this option? 

Assuming tasks (async calling) will need to align with MCP, I'd expect Claude SDK to be the closest, good reference implementation

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+This is the approach taken by the [Strands Durability Plugin (SARK)](https://sark-docs.beta.harmony.a2z.com/sdo-strands-durability-plugin/developer-guide/async-execution). SARK's interrupt-based session release is more applicable to the durable task phase than to the ephemeral mechanism proposed here.
+
+| Pros | Cons |
+|------|------|
+| Zero model complexity — no new tools, no async semantics to learn | Model cannot dispatch multiple tasks concurrently |
+| Releases compute during long waits — no idle GPU/container | Model cannot reason about partial results as they arrive |
+| No risk of hallucinated results or polling loops | Only suited for single long-running tasks, not parallel workloads |
+| Works with any existing tool without modification | Requires platform-level re-invocation infrastructure |
+
+**This approach is complementary, not competing.** `background()` keeps the model productive while multiple tasks run. Model-unaware async releases compute when the model has nothing else to do. In a managed deployment, both compose: the model dispatches background tasks, and the platform releases compute while waiting for results.
```
zastrowm, line 167: @JackYPCOnline, @pgrayy this sounds like the invoke checkpoint stuff maybe?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+---
+
+## Extension to Agent Harness
+
+Background Tasks as proposed are ephemeral: in-process, tracked in memory, scoped to the agent's lifetime. If the agent process dies, the tasks die with it. This is acceptable for interactive sessions and short-lived agentic workflows, but does not scale for long-running production tasks.
```
zastrowm, line 173: So no integration with session-manager until we extend it further?  I *think* that's fine, just trying to clarify

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+| Explicit polling (LangChain) | Zero hallucination risk — model actively retrieves results | Token overhead from repeated status checks; wasted model calls when results aren't ready |
+| Status + auto-notification (Claude Agent SDK) | Clear "async_launched" signal reduces hallucination; model can poll or wait | More complex — requires both notification infrastructure and polling tools |
+
+### Model-unaware async
+
+The following is not an alternative to Background Tasks but a complementary approach for a different use case.
+
+The model calls a tool normally. Behind the scenes, the framework intercepts the result, releases the agent process, and waits for the external work to complete. The agent is re-invoked when ready, and the result is injected as if the tool ran synchronously. The model never knows anything was async.
+
+This is the approach taken by the [Strands Durability Plugin (SARK)](https://sark-docs.beta.harmony.a2z.com/sdo-strands-durability-plugin/developer-guide/async-execution). SARK's interrupt-based session release is more applicable to the durable task phase than to the ephemeral mechanism proposed here.
```
mkmeral, line 158: cc @JackYPCOnline  are you aware of this, as part of durability work?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Ephemeral background tasks
+
+The base-level mechanism. Tracked in memory, scoped to the agent's lifetime, no infrastructure required. Ships as part of this proposal.
+
+### Durable background tasks
+
+The model calls the same `background()` tool. What changes is the infrastructure underneath. Tasks run outside the agent process and survive process death. The agent can resume from where it left off, and task metadata persists across sessions.
+
+Durable background tasks enable fire-and-forget: the agent can dispatch work, exit, and pick up results on its next invocation. They also open the door to ambient, long-running processes — background agents that persist across sessions, react to events, and operate alongside the main agent over extended timeframes.
```
zastrowm, line 185: I think there's a middleground here for "durable-tools" of a sort? E.g. tools could-opt into this by returning a token that represents the tool response, and that token is stored in session-management; upon resume the token is given back to the tool to check it's status and it effectively resumes the task

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+  tools: [quickLookup],
+  backgroundTools: [searchWeb, analyzeData],
+})
+
+// Model calls searchWeb as usual — the SDK backgrounds it automatically
+```
+
+| Pros | Cons |
+|------|------|
+| Developer controls which tools are safe to background (stateful tools, side effects) | Developer decides what's backgroundable, not the model — no per-conversation flexibility |
```
JackYPCOnline, line 102: If user defined tools, that makes sense. What about sdk vended tools.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+### Unrelated tasks block each other
+
+If task A finishes in 2 seconds and task B takes 30, the model cannot act on A's result until B also completes — even if A and B have nothing to do with each other. A completed result should be actionable the moment it arrives.
+
+---
+
+## Proposal
+
+Background Tasks give the model control over *when* work runs, not just *what*. A single `background()` tool lets the model dispatch work, keep reasoning, and react to results as they arrive:
```
pgrayy, line 36: How does the model decide when to call a tool in background mode?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+  [tool_result] background → { taskId: "t-1", status: "working" }
+  [tool_result] calculateMetrics → "Revenue: $4.2M, Growth: 18%"
+
+Turn 3:
+  [assistant]  "Metrics show $4.2M revenue with 18% growth. Search results
+               are on the way — I'll incorporate them when they arrive."
+
+  ← search_web completes →
+
+  [message]    { tool: "search_web", taskId: "t-1", status: "completed",
```
mkmeral, line 240: nit: this is the start of Turn 4

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+|------|------|
+| Developer controls which tools are safe to background (stateful tools, side effects) | Developer decides what's backgroundable, not the model — no per-conversation flexibility |
+| No additional tool in the model's schema | A tool is always background or always foreground — can't adapt at runtime |
+| Model can't misuse it — only pre-approved tools run in the background | Two tool lists to maintain; overlap between them creates ambiguity |
+| Simpler model behavior — fewer failure modes (no hallucinated results, no polling loops) | Requires system prompt augmentation so the model understands async results |
+| | Conflicts with the core Strands principle of letting the model drive decisions |
+
+### Per-tool config
+
+Instead of a meta-tool or a separate list, the background property lives on each tool definition. The model still calls tools normally — the SDK checks the per-tool config and backgrounds the call automatically.
```
zastrowm, line 110: Does it make sense to allow tools to opt-in to be background tasks only?  E.g. I could imagine a scenario where a tool always returns an ephemeral/durable response and so instead of allowing the agent to invoke it synchronously it forces it to be background?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+```typescript
+const agent = new Agent({
+  tools: [
+    quickLookup,
+    { tool: searchWeb, background: true },
```
JackYPCOnline, line 116: do we need to define what tools are background-able?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+The model reasons about the foreground result immediately (Turn 3), then incorporates the background result when it arrives (Turn 4). The tagged message lets the model correlate the result to its original dispatch.
+
+---
+
+## Appendix B: Prior Art
+
+Several implementations validate the core pattern of async dispatch for Strands agents:
+
+**[async-agentic-tools](https://github.com/mikegc-aws/async-agentic-tools)** (mikegc-aws) — ~320 lines of Python. A `@tool_async` decorator wraps any Strands tool, dispatches it to a thread pool, and returns a task ID immediately. An `AsyncAgent` wrapper handles result delivery via callbacks. Proves the pattern works today with minimal code, though the developer decides which tools are async (decorator-based), not the model.
```
mkmeral, line 256: but like, does it work? just because someone wrote the code, doesn't mean it works 😅  (especially with all the AI slop), is it proven?

designs/0009-background-tasks.md
pgrayy, line 36: Let's say I have a tool that calls an API in background mode. Could this be setup so the model calls this tool as normal but the tool returns with a response id that should be passed back in when the response is available?

This fits in line with OpenAI's model background mode. When you call the model in background mode, you get a response id that you pass back in a request to retrieve the response when it is done generating.

Through this approach, you wouldn't need a special background tool for the model. However, each tool implementation would then be responsible for exposing background functionality if it supports it.

designs/0009-background-tasks.md
JackYPCOnline, line 158: hmm, interesting

designs/0009-background-tasks.md
lizradway: +1 similarly, if we were to introduce polling, when would the model poll?

designs/0009-background-tasks.md
mkmeral: > More complex — requires both notification infrastructure and polling tools 

is that a problem? claude is writing the code anyways :p 

I like this approach better compared to what's recommended tbh. The main reason is, the agent might be in the middle of a task, and your tool result can be huge. Dumping it can confuse the agent. In this approach, we leave a notification (instead of the whole tool result), and let the agent get to it, when they can.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+*The model can decide what to do, but not when. New work always waits for old work to finish, even when the two are unrelated.*
+
+### The agent is blocked while tools are running
+
+A 30-second API call means the agent idles for 30 seconds. A 15-minute MCP tool running a deep research pipeline blocks the agent for the full 15 minutes — whether or not the result is needed before moving on. The agent has no way to make progress on anything else while it waits.
+
+### The model has no control over concurrency
+
+Every concurrency option in Strands requires developer configuration. If the model determines mid-reasoning that two tasks are independent, it has no way to dispatch them and keep reasoning while they run. ConcurrentToolExecutor helps — it drops latency from sum(tools) to max(tools) — but the developer decides what runs concurrently, not the model, and the agent still blocks until the batch completes.
```
lizradway, line 26: are we planning on limiting a maximum number of tasks that can run in the background concurrently? mainly wondering how this scales? any ideas how many background tasks can/should be run at once and when this becomes no longer beneficial?

designs/0009-background-tasks.md
mkmeral: > when would the model poll?

depends on what you tell the model. if you don't tell anything, it starts and keeps polling, causing context bloat

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+| Remote agent (A2A) | `background({ ..., tool: "remote_agent", args: { input: "..." } })` |
+| Dynamic agent (use_agent) | `background({ ..., tool: "use_agent", args: { prompt: "...", tools: [...] } })` |
+| Any callable | `background({ ..., tool: "x", args: { ... } })` |
+
+The model keeps reasoning and dispatching more work as needed. As results come in, they are delivered as tagged messages appended to the conversation, not as `tool_result` blocks. The dispatch returns a task ID as a normal `tool_result` (satisfying the provider API's synchronous pairing requirement), and the actual result arrives later as a separate message tagged with the originating tool and task ID. The model sees them at the start of its next turn. The tool definition instructs the model to continue working without fabricating results for dispatched tasks.
+
+Background results are subject to the same context window limits and compaction behavior as any other conversation content. Proactive strategies — structured output on sub-agents to reduce result size, batching closely-completing results into a single injection, or queuing results when the context is near capacity — can mitigate this further.
```
pgrayy, line 58: What happens if we have context bloat or context summarization while waiting for a tool result. Are there plans to help mitigate confusing the model with a tool result that arrives late? How do we help the model not forgot why it wanted the tool result in the first place?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+The path from ephemeral to *durable* requires three things: isolated execution (so tasks outlive the agent process), persistent state (so the agent can resume), and an external task store (so task metadata survives). This is what an agent harness provides.
+
+### Ephemeral background tasks
+
+The base-level mechanism. Tracked in memory, scoped to the agent's lifetime, no infrastructure required. Ships as part of this proposal.
+
+### Durable background tasks
+
+The model calls the same `background()` tool. What changes is the infrastructure underneath. Tasks run outside the agent process and survive process death. The agent can resume from where it left off, and task metadata persists across sessions.
```
JackYPCOnline, line 183: I am not sure, but this actually has conflict with agent controlled event loop idea..? I think we can add durability in dispatch path as well.

designs/0009-background-tasks.md
lizradway: im also wondering why we didn't go with this option. "more complex" from a development/implementation perspective or maintainability and customer use perspective? if this produces better results, i'd rather go with this.


might be worth running a few test scripts with each access method to get some stats on success rates of each.

designs/0009-background-tasks.md
mkmeral: I think it makes sense. I'd expect tool developer to have some control here. That said, would it make sense to start this under experimental, and once we settle, we can add more config options and move to main? 🤔

designs/0009-background-tasks.md
pgrayy: https://developers.openai.com/api/docs/guides/background
^ Info on model background mode.

designs/0009-background-tasks.md
mkmeral: On the same note, how do we handle agent stream, when we re-invoke the agent from background tool result? Is hte agent invoked as part of previous `agent.stream_async` call, or is this a separate async method? in which case we need a way to actually stream the agent response

(I think solution is callback handler, but not sure if that's the pattern we like, i guess?)

designs/0009-background-tasks.md
lizradway: will these tool results be lost if context is compressed mid-task?

designs/0009-background-tasks.md
JackYPCOnline: it is more like very initial version: dispatch tool call [model call, tool calls in our example], wrap the call in a durable provider service.

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+A 30-second API call means the agent idles for 30 seconds. A 15-minute MCP tool running a deep research pipeline blocks the agent for the full 15 minutes — whether or not the result is needed before moving on. The agent has no way to make progress on anything else while it waits.
+
+### The model has no control over concurrency
+
+Every concurrency option in Strands requires developer configuration. If the model determines mid-reasoning that two tasks are independent, it has no way to dispatch them and keep reasoning while they run. ConcurrentToolExecutor helps — it drops latency from sum(tools) to max(tools) — but the developer decides what runs concurrently, not the model, and the agent still blocks until the batch completes.
+
+### Unrelated tasks block each other
+
+If task A finishes in 2 seconds and task B takes 30, the model cannot act on A's result until B also completes — even if A and B have nothing to do with each other. A completed result should be actionable the moment it arrives.
```
notowen333, line 30: aside: This makes me think whether there should be guidance to include estimated latencies in tool descriptions

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+---
+
+## Proposal
+
+Background Tasks give the model control over *when* work runs, not just *what*. A single `background()` tool lets the model dispatch work, keep reasoning, and react to results as they arrive:
+
+| Action | Example | Description |
+|--------|---------|-------------|
+| dispatch | `background({ action: "dispatch", tool: "search_web", args: { query: "..." } })` | Dispatch work, receive a task ID immediately |
```
notowen333, line 40: Strands SDK will implicitly create these 3 tools in the registry when the user creates a background tool?

designs/0009-background-tasks.md
```diff
@@ -0,0 +1,301 @@
...
+
+| Pros | Cons |
+|------|------|
+| Each tool has a focused schema — simpler for the model per-call | Three tools in the registry instead of one |
+| Matches the MCP Tasks protocol shape (`tasks/get`, `tasks/cancel`) | Model must discover and learn three tools instead of one |
+| No `action` parameter to misuse | More surface area to maintain |
+
+Either shape works. We propose the single-tool approach because it keeps the registry minimal and the model only needs to learn one tool name.
+
+### Result delivery strategy
```
mkmeral, line 141: Is this something tool manually does? Or do we have a generic way of doing it? or will we have it?

designs/0009-background-tasks.md
notowen333, line 68: I'm curious to understand if we have a sense of which domain use cases are best fit for each background/concurrency approach.

i.e. if I can make agents as tools or background agents as tools or a graph with concurrent edges etc how can we guide users towards the best fit?

designs/0009-background-tasks.md
pgrayy: Also, could be helpful to think about what happens if users filter out tool uses/tool results immediately after model responds to a tool result. Not sure if this happens in practice but I do wonder if someone has tried it for token saving strategies especially when caching isn't available. So the flow is:
1. "User question"
2. "Tool use"
3. "Tool result"
4. "Model response"

Under this flow, tool use and tool result can be treated as metadata and so can be filtered out. The critical context is now captured in "user question" and "model response". This means though the model loses the "dispatch" context when running a tool in background mode.

pgrayy: Also worth thinking about what happens when running a model in stateful mode. We don't control message history under those circumstances.

PR #795, opened 2026-04-28
Homepage update

ryanycoleman:
## Description
Staging some updates to the website including copy edits and a language switcher that includes Typescript code samples on the homepage. 

## Checklist

----


src/components/landing/CodeBlock.astro
```diff
@@ -10,15 +10,17 @@ interface Props {
...
 
-const { code, lang = 'python', filename } = Astro.props
+const { code, lang = 'python', filename, badge } = Astro.props
 ---
 
 <div class="code-block" data-code={code}>
   {filename && (
     <div class="code-header">
       <span class="code-filename">{filename}</span>
+      {badge && <span class="code-badge">{badge}</span>}
```
zastrowm, line 23: This isn't really readable in dark mode:

Image

src/components/landing/LangCodeBlock.astro
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  [data-lang-code="Python"] {
+    display: block;
+  }
+
+</style>
+
+<script>
+  const STORAGE_KEY = 'starlight-synced-tabs__jarkqt'
```
zastrowm, line 54: As a follow-up can we extract this to a common location?

src/components/landing/StatsSection.astro
```diff
@@ -0,0 +1,115 @@
...
+
+<section class="stats-section">
+  <div class="stats-inner">
+    <div class="proof-stats">
+      <div class="stat-primary">
+        <span class="stat-number">100%</span>
+        <span class="stat-label">task accuracy with steering</span>
+      </div>
+      <div class="stat-comparison">
+        Prompt-only agents scored <strong>82.5%</strong>. Hard-coded workflows scored <strong>80.8%</strong>. Steered agents recovered from every mistake.
```
zastrowm: The emphasis here makes it hard to identify that these are separate sentences:

Image

I had to read it a couple times to realize it wasn't saying "80.8% Steered agents recovered" for example

src/components/landing/QuoteCarousel.astro
```diff
@@ -0,0 +1,254 @@
...
+        return (
+          <div class="quote-card">
+            <blockquote class="quote-text">
+              <Content />
+            </blockquote>
+            <div class="quote-author">
+              <div class="author-details">
+                <span class="author-name">{testimonial.name}</span>
+                {testimonial.title && <span class="author-title">{testimonial.title}</span>}
+                {testimonial.link && <a href={testimonial.link} class="quote-read-more" target="_blank" rel="noopener">Read more &rarr;</a>}
```
zastrowm, line 36: Possible as a follow-up, put read-more below the quote instead; otherwise we get aa mix of authors that are 3 lines and those that are 2

Image

src/components/landing/QuoteCarousel.astro
```diff
@@ -0,0 +1,254 @@
...
+    /* Clamp to ~4 lines by default */
+    display: -webkit-box;
+    -webkit-line-clamp: 4;
+    -webkit-box-orient: vertical;
+    overflow: hidden;
+    transition: max-height 0.3s ease;
+  }
+
+  /* Expanded state (toggled by JS on hover) */
+  .quote-text.expanded {
```
zastrowm: We should expand only a single quote here and have it overlap/overlay content below instead of having the height affect the entire document - it pushes all content down on hover and then back up after hover - sometimes even as you scroll.

src/components/landing/UseCasesSection.astro
```diff
@@ -0,0 +1,324 @@
...
+    <div class="use-case">
+      <div class="use-case-text">
+        <h3 class="use-case-title">Build AI assistants</h3>
+        <p class="use-case-description">Ground agents in your knowledge base via MCP. Context management keeps long conversations in bounds. Interrupts pause for human approval before sensitive actions.</p>
+      </div>
+      <div class="use-case-code">
+        <LangCodeBlock pythonCode={assistantCode} pythonFilename="support_agent.py" tsCode={assistantTs} tsFilename="support_agent.ts" />
+      </div>
+    </div>
+    <div class="use-case">
```
zastrowm, line 237: Not a blocker, but I still think this layout would be more readable:

```
Description text
    indented code example



                          Description text
left-aligned code
with right-margin
```

Right now the code feels a bit smooshed against the descriptions

zastrowm (changes requested):

zastrowm (approved):

PR #805, opened 2026-05-01
fix: default homepage code snippets to typescript

lizradway:
## Description

Fixes a first-visit bug where the homepage language toggle (LanguageToggle) defaults to TypeScript but the code snippets (LangCodeBlock) default to Python, causing them to disagree when localStorage is empty.

Three changes:
- **Align JS fallback** in LangCodeBlock from `'Python'` to `'TypeScript'` to match LanguageToggle
- **Flip CSS defaults** in LangCodeBlock to show TypeScript and hide Python before JS runs (scoped to paired blocks so Python-only examples like steering/guardrails remain visible)
- **Write default to localStorage on init** in LanguageToggle so the preference is available to all consumers immediately, not just after a click

## Related Issues

N/A

## Type of Change

- Bug fix

## Checklist

----

### DIFF
--- a/src/components/LanguageToggle.astro
+++ b/src/components/LanguageToggle.astro
@@ -216,6 +216,7 @@
 
     // Restore on page load
     var current = getPreference();
+    setPreference(current);
     if (document.readyState === 'loading') {
       document.addEventListener('DOMContentLoaded', function () { syncAll(current); });
     } else {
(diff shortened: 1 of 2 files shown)


ryanycoleman (approved):
works great on mobile (fresh browser)

PR #806, opened 2026-05-01
docs: add proactive compression docs

opieter-aws:
## Description

Add docs for proactive compression and token counting features

## Related Issues

https://github.com/strands-agents/sdk-python/pull/2239

## Type of Change

- New content

## Checklist

----

### DIFF
--- a/src/content/docs/user-guide/concepts/agents/conversation-management.mdx
+++ b/src/content/docs/user-guide/concepts/agents/conversation-management.mdx
@@ -96,6 +96,7 @@ Key features of the `SlidingWindowConversationManager`:
 - **Overflow Trimming**: In the case of a context window overflow, it will trim the oldest messages from history until the request fits in the models context window.
 - **Configurable Tool Result Truncation**: Enable / disable truncation of tool results when the message exceeds context window limits. When `should_truncate_results=True` (default), large results are truncated with a placeholder message. When `False`, full results are preserved but more historical messages may be removed. For a proactive alternative that preserves full content externally, see the [Context Offloader](../plugins/context-offloader) plugin.
 - **Per-Turn Management**: Optionally apply context management proactively during the agent loop execution, not just at the end.
+- **Proactive Compression** (TypeScript): Pass `proactiveCompression: true` or `proactiveCompression: { compressionThreshold: 0.7 }` to trigger context reduction before the model call when projected input tokens exceed a configurable threshold. See [Proactive Context Compression](#proactive-context-compression).
 
 **Per-Turn Management**:
 
@@ -160,6 +161,7 @@ Configuration parameters:
 - **`summaryRatio`** (number, default: 0.3): Ratio of messages to summarize when reducing context (clamped between 0.1 and 0.8)
 - **`preserveRecentMessages`** (number, default: 10): Minimum number of recent messages to always keep
 - **`summarizationSystemPrompt`** (string, optional): Custom system prompt for summarization. If not provided, uses a default prompt that creates structured bullet-point summaries focusing on key topics, tools used, and technical information in third-person format.
+- **`proactiveCompression`** (`boolean | { compressionThreshold: number }`, optional): Enable proactive context compression before the model call. Pass `true` for the default 0.7 threshold, or an object with a custom threshold. See [Proactive Context Compression](#proactive-context-compression).
 
 </Tab>
 </Tabs>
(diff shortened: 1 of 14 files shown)


src/content/docs/user-guide/concepts/agents/conversation-management.ts
```diff
@@ -121,3 +122,34 @@ Format as bullet points without conversational language.
   })
   // --8<-- [end:summarizing_conversation_manager_system_prompt]
 }
+
+async function proactiveSlidingWindow() {
+  // --8<-- [start:proactive_sliding_window]
+  const agent = new Agent({
+    model: new BedrockModel({
+      modelId: 'anthropic.claude-sonnet-4-20250514-v1:0',
+      // contextWindowLimit is auto-populated from the
```
lizradway: this info is duplicated in the section below. i think it makes sense to keep it there?

src/content/docs/user-guide/concepts/agents/conversation-management.ts
```diff
@@ -121,3 +122,34 @@ Format as bullet points without conversational language.
   })
   // --8<-- [end:summarizing_conversation_manager_system_prompt]
 }
+
+async function proactiveSlidingWindow() {
```
lizradway, line 125: might be nice to include an example where the user does not set the threshold so we can see the default to `0.7`

src/content/docs/user-guide/concepts/agents/conversation-management.mdx
```diff
@@ -310,6 +312,81 @@ agent = Agent(
...
+The threshold check requires the model's context window size. The SDK auto-populates `contextWindowLimit` from built-in lookup tables for known models (Anthropic, Bedrock, OpenAI, Google Gemini). You can override it manually for models not in the lookup table:
+
+```typescript
+const model = new BedrockModel({
+  modelId: 'my-custom-model',
+  contextWindowLimit: 128_000,
+})
+```
+
+If `contextWindowLimit` is not set and the model ID is not in the built-in lookup table, the SDK falls back to a default of 200,000 tokens.
```
lizradway: Maybe combine this into the caution below? I like that this is directly noted as warning so that we can cite it if we get any bugs/pushback lol.

I think that the caution section can be simplified probably, since the after effects if limit is incorrect are pretty self explanatory imo (I'd hope) ie.  what happens if we compress too early or too late.

src/content/docs/user-guide/concepts/agents/conversation-management.mdx
```diff
@@ -310,6 +312,81 @@ agent = Agent(
 - **Flexible Configuration**: Customize summarization behavior through various parameters
 - **Fallback Safety**: Handles summarization failures gracefully
 
+## Proactive Context Compression
```
lizradway, line 315: Might be worth moving this below "Creating a conversation manager"

src/content/docs/user-guide/concepts/agents/conversation-management.mdx
```diff
@@ -310,6 +312,81 @@ agent = Agent(
...
+
+- **Actual limit is smaller** (e.g., 8k or 32k): Compression triggers too late. The model will hit context overflow errors before proactive compression is triggered, falling back fully to reactive overflow recovery.
+- **Actual limit is larger** (e.g., 1M): Compression triggers too early. The agent proactively trims context well before the model's capacity is reached.
+:::
+
+### Token Estimation
+
+The agent estimates input tokens using a the following strategy:
+
+1. **Known baseline**: Reads `inputTokens + outputTokens` from the last assistant message's metadata
```
lizradway: Might be nice to point out where these live on the agent result interface output?

src/content/docs/user-guide/concepts/agents/conversation-management.mdx
```diff
@@ -310,6 +312,81 @@ agent = Agent(
...
+2. **Delta estimation**: Only estimates tokens for new messages added since that assistant message using the model's `countTokens()` method
+3. **Cold start fallback**: When no prior usage metadata exists (first call or after session restore without metadata), estimates all messages via `countTokens()`
+
+The `countTokens()` method uses a character-based heuristic to estimate token count by default (characters ÷ 4 for text, characters ÷ 2 for JSON). Some model providers support native token counting APIs for exact counts, which can be enabled on the model. See the Token Counting section on each provider's page for details and instructions:
```
lizradway, line 437: >See the Token Counting section  

Add link?

src/content/docs/user-guide/concepts/agents/conversation-management.mdx
lizradway, line 1: Python code examples?

src/content/docs/user-guide/concepts/agents/conversation-management.mdx
```diff
@@ -310,6 +312,81 @@ agent = Agent(
...
+    reduce()  // proactively compress context
+```
+
+Each conversation manager uses the same reduction logic for proactive compression as reactive overflow recovery. Proactive compression is best-effort only, so if `reduce()` throws or returns `false`, the error is swallowed and the model call proceeds normally.
+
+Because `BeforeModelCallEvent` triggers before every model call including calls within a tool-use cycle, this provides automatic in-loop compression. If an agent makes five tool calls in a single invocation and context grows past the threshold between calls three and four, compression triggers before call four.
+
+### Context Window Limit
+
+The threshold check requires the model's context window size. The SDK auto-populates `contextWindowLimit` from built-in lookup tables for known models (Anthropic, Bedrock, OpenAI, Google Gemini). You can override it manually for models not in the lookup table:
```
lizradway: nit: (Anthropic, Bedrock, OpenAI, Google Gemini) imo isn't a necessary specification. i dont really want to update this doc in the future if we change context window table lol. maybe just link to the lookup table?

src/content/docs/user-guide/concepts/model-providers/amazon-bedrock.mdx
```diff
@@ -917,6 +917,43 @@ print(f"Rating: {result.rating}")
 </Tab>
 </Tabs>
 
+### Token Counting
+
+Token counting is used by [proactive context compression](/docs/user-guide/concepts/agents/conversation-management/#proactive-context-compression) to estimate input tokens before each model call.
```
lizradway: might be nice to just cite "Token counting is used by context management strategies" rather than proactive compression specifically since this is also used by offloading + will be used by a bunch of stuff in the future.

src/content/docs/user-guide/concepts/model-providers/openai-responses.mdx
lizradway, line 1: out of curiosity: is this only available in responses, not messages api?

opieter-aws: Correct, only the responses api supports the endpoint

src/content/docs/user-guide/concepts/agents/conversation-management.mdx
opieter-aws: The links are listed below. They're individual sections for each model provider.

src/content/docs/user-guide/concepts/agents/conversation-management.mdx
opieter-aws: The section `Creating a conversation manager` is about creating a custom conversation manager. To me it's clearer to present that only after the documentation of all built-in features.

lizradway (dismissed):

lizradway (approved):

PR #831, opened 2026-05-11
feat(designs): context built in strategy

lizradway:
## Type of Change

- New content
- Content update/revision
- Typo/formatting fix
- Bug fix
- Other (please describe):

## Checklist

----

### DIFF
--- a/designs/0010-context-strategy.md
+++ b/designs/0010-context-strategy.md
@@ -1,0 +1,602 @@
+# Design: Context Management Presets
+
+- [1. Problem Statement](#1-problem-statement)
+- [2. The Cache Hierarchy](#2-the-cache-hierarchy)
+- [3. Presets](#3-presets)
+- [4. TypeScript v1](#4-typescript-v1)
+- [5. TypeScript v2](#5-typescript-v2)
+- [6. Anticipated Questions](#6-anticipated-questions)
+- Appendix A: Code Examples
+- Appendix B: Alternatives Considered
+- Appendix C: Composition Edge Cases
+- Appendix D: Plugin & Strategy Details
+- Appendix E: L1 Storage Model
+
+---
+
+## 1. Problem Statement
+
+Every non-trivial agent eventually hits context limits. Strands already provides the building blocks to handle this — externalization, compression, conversation managers, hooks — but today these are independent extension points that users must discover, configure, and compose themselves. The SDK covers the 20% case (power users who want full control) but not the 80% case (developers who want context management to just work).
+
+As context management ships more capabilities, this gap widens. Each feature is another plugin or tool to wire up. The Strands [tenets](https://github.com/strands-agents/docs/blob/main/team/TENETS.md) call for **the obvious path to be the happy path** and for **simple things to be simple**. Context management should be a one-liner, not a composition exercise.
+
+This design proposes opinionated defaults via a single `contextManager` parameter on `Agent` — backed by the same extension points that power users already configure manually, but with sensible choices pre-made for everyone else.
+
+---
+
+## 2. The Cache Hierarchy
+
+> **Note:** This doc covers L0 ↔ L1 (within-session context management). L2 and anything memory-related is out of scope — see Thomas's Memory primitive design. References to memory in this doc are included only for background.
+
+Context management maps to a three-tier cache hierarchy. Every operation is movement between tiers.
+
+| Tier | What it is | Access cost | Backed by |
+|------|-----------|-------------|-----------|
+| **L0 — Context window** | `agent.messages` — what the model sees on every request. May contain summaries where originals were evicted. | Zero (already in context) | In-memory message list |
+| **L1 — Session history** | Append-only log of messages. Only written when there's a consumer (agentic mode or `memoryManager`). Messages are batch-written to L1 before each summarization. See Appendix E for storage internals. | Low (retrieval tool call) | `ContextManager` storage (`FileStorage`, `S3Storage`, etc.) |
+| **L2 — Long-term memory** | Cross-session knowledge — archival memory, semantic search index, learned facts. Persists beyond the current session. | Higher (query + load from persistent storage) | Memory primitive, vector store, managed memory services |
+
+L2 (long-term memory) is a separate primitive (`memoryManager`) — out of scope for this document. From this doc's perspective, L1 is append-only and immutable for the session lifetime. The tool result cache is separate — short-lived and auto-evicting, not immutable.
+
+See Appendix E for L1 storage internals.
+
+---
+
+## 3. Presets
+
+The `contextManager` parameter accepts a **strategy** that determines who controls L0 — what stays in the context window and when things get compressed to L1.
+
+Both strategies share the same infrastructure. When context pressure builds, L0 is compressed. Oversized tool results are cached separately, and `retrieveToolResult` is always available so the agent can recover truncated outputs without reasoning about context.
+
+The difference: in **auto**, the agent interacts with content. In **agentic**, the agent reasons about its own context.
+
+The logic lives in plugins, not tools. Tools are agent-facing wrappers around plugin methods. In auto mode, the framework calls plugin methods via hooks (e.g., `ContextCompression.compress()` fires on threshold). In agentic mode, the same hooks still fire as a safety net, but tools are also registered so the agent can invoke the same methods proactively. The strategy controls which tools are exposed — not what logic exists.
+
+| | `"auto"` | `"agentic"` |
+|--|----------|-------------|
+| Framework compresses when threshold hit | Yes | Yes |
+| Framework caches oversized tool results | Yes | Yes |
+| Agent can recover truncated tool output | Yes (`retrieveToolResult`) | Yes |
+| Agent can trigger compression | No | Yes (`compressContext`) |
+| Agent can protect messages from eviction | No | Yes (`pinMessage`) |
+| Agent can read from L1 | No | Yes (`getHistory`) |
+| Agent can search L1 | No | Yes (`searchHistory`) |
+| Agent can spawn context-aware children | No | Yes (`delegateWithContext`) |
+| Agent can check its context budget | No | Yes (`getContextBudget`) |
+
+Users who don't want presets can build custom context management using the same infrastructure. Token tracking, message metadata, token estimation, and context limit primitives together give users everything they need to write their own `BeforeModelCallEvent` hooks with custom policies. The presets are opinionated compositions of these primitives, not the only way to use them.
+
+### `"auto"` — The framework decides
+
+The framework manages L0 transparently. When context pressure builds, messages are summarized or dropped in L0. No L1 writes happen in auto mode — L1 is only written when there's a consumer (agentic mode or `memoryManager`). The agent doesn't know context management is happening — it just never hits a wall. The only tool exposed is `retrieveToolResult` for recovering truncated tool outputs.
+
+**Use cases:** Beginners who don't want to think about context yet. Production deployments where predictability matters more than autonomy. Cost-sensitive workloads at scale. Multi-agent systems where child agents need lightweight, hands-off management. Anyone who wants it to just not break.
+
+### `"agentic"` — The agent decides
+
+> **Note:** `"agentic"` is experimental. It depends on research into whether models effectively use context management tools. `"auto"` ships first and is the primary focus.
+
+Everything `"auto"` does still happens — the agent *also* gets tools to actively manage its own L0. It can decide when to compress, what to protect from eviction, and browse evicted messages in L1.
+
+**Use cases:** Research agents and coding assistants that benefit from self-awareness about their context state. Long-running autonomous agents. Exploratory development where agent autonomy is the point. Parent agents in multi-agent orchestrations.
+
+### Growth and ownership
+
+More strategies may be added as context management capabilities evolve. Per the [pay-for-play tenet](https://github.com/strands-agents/docs/blob/main/team/TENETS.md), the Strands team owns these presets and reserves the right to make breaking changes to them — what a strategy resolves to underneath (which plugins, which defaults) can change between versions. Users who need stability should configure plugins directly rather than relying on preset internals.
+
+Defaults for all preset configurations (thresholds, strategies, token limits) should be informed by benchmarking before shipping. No default should be set based on intuition alone.
+
+---
+
+## 4. TypeScript v1
+
+Ships `contextManager: "auto"` as an opt-in parameter.
+
+### What ships
+
+```typescript
+// Minimal — defaults to SandboxStorage
+const agent = new Agent({ contextManager: "auto" });
+
+// Explicit storage
+const agent = new Agent({
+  contextManager: { strategy: "auto", storage: new S3Storage({ bucket: "my-session" }) },
+});
+
+// Disable offloading (compression only)
+const agent = new Agent({
+  contextManager: { strategy: "auto", contextOffloader: false },
+});
+
+// Custom offloader threshold
+const agent = new Agent({
+  contextManager: { strategy: "auto", contextOffloader: { threshold: 5000 } },
+});
+```
+
+`contextManager` accepts a string shorthand or a config object in v1. The Agent resolves the input directly into plugins and tools — no intermediate class. The `ContextManager` class (owning token estimation, budget, cursors) is introduced in v2 for power users who need lifecycle hooks or direct storage access.
+
+
+In v1, the default is `undefined` (no context management). This avoids surprises on upgrade and gives us time to prove the behavior works before making it default.
+
+Under the hood, this wires up:
+
+| Component | What it does |
+|-----------|-------------|
+| **`ContextOffloader`** | Oversized tool results cached separately (short-lived, auto-evicting) via `AfterToolCallEvent` hook. Agent sees a truncated preview. Provides `retrieveToolResult` tool for on-demand access. Renamed to `ToolResultCache` in v2 (config key changes from `contextOffloader` to `toolResultCache`). |
+| **Proactive compression on conversation manager** | `BeforeModelCallEvent` hook checks L0 token usage against threshold. When exceeded, compresses older messages in L0. |
+| **Message protection** | Position-based pinning (`protectedMessages = 1` by default) ensures the task prompt is never evicted. |
+
+### Coexistence with `conversationManager`
+
+In v1, the user's conversation manager is respected. The preset configures proactive compression *on* whichever conversation manager the user provides (or the default):
+
+```typescript
+// Uses default conversation manager with proactive compression
+const agent = new Agent({ contextManager: "auto" });
+
+// Uses user's conversation manager, still gets proactive compression
+const agent = new Agent({
+  contextManager: "auto",
+  conversationManager: new SummarizingConversationManager(),
+});
+```
+
+---
+
+## 5. TypeScript v2
+
+Three changes:
+1. **`"auto"` becomes the default** — every agent gets context management unless opted out
+2. **`"agentic"` ships** — agent-in-the-loop context management
+3. **`conversationManager` is deprecated** — its compression responsibility moves to `ContextCompression`
+
+### Plugin decomposition
+
+Each plugin has a single cohesive responsibility. `ContextManager` is not a plugin — it's the top-level class that the config resolves to. It owns shared infrastructure (storage, token estimation, budget tracking, cursors) and composes the plugins below. Plugins read from `ContextManager` rather than depending on each other.
+
+| Component | Role | Domain | Tools | Mode |
+|-----------|------|--------|-------|------|
+| **`ContextManager`** | Config resolver + shared infrastructure | Storage, token estimation, budget, cursors | `getContextBudget` (agentic) | Always |
+| **`ToolResultCache`** | Plugin | Cache oversized tool results | `retrieveToolResult` | Both |
+| **`ContextCompression`** | Plugin | L0 writes — compression, pinning | `compressContext`, `pinMessage` (agentic) | Both |
+| **`ContextNavigation`** | Plugin | Read from L1 (session history) | `getHistory`, `searchHistory` | Agentic only |
+| **`ContextDelegation`** | Plugin | Context-aware child spawning | `delegateWithContext` | Agentic only |
+
+v2 introduces `OnContextOverflowEvent` — a new lifecycle event fired on context length errors, replacing the opaque retry logic in conversation managers. The SDK already normalizes these errors across all providers into `ContextWindowOverflowError` — this event just exposes that to plugins.
+
+### `conversationManager` deprecation
+
+In v2, `ContextCompression` takes over compression and L1 writing. Nothing left for `conversationManager`.
+
+**Deprecation path:**
+1. **v1 (coexist):** Both live side by side. No conflict.
+2. **v2 (deprecate):** Deprecation warning emitted. `contextManager` takes precedence if both are set.
+3. **v3 (remove):** `conversationManager` removed from the constructor type.
+
+---
+
+## 6. Anticipated Questions
+
+**Q: Why isn't `"auto"` the default in v1?**
+
+We need to prove the behavior first. v1 is opt-in so we can validate with early adopters. v2 makes it the default once we're confident it works reliably.
+
+**Q: What happens when users upgrade from v1 to v2?**
+
+Two changes: (1) `"auto"` becomes the default — agents that didn't set `contextManager` now get it automatically. Users who don't want this can set `contextManager: false`. (2) Internals upgrade — `ToolResultCache` replaces `ContextOffloader`, `ContextCompression` replaces conversation manager compression — same external behavior.
+
+**Q: What if the user provides their own plugin?**
+
+Deduplication detects by type and skips the preset's version. User's `ToolResultCache`, `ContextCompression`, or `ContextNavigation` (v2) wins.
+
+**Q: Why a config object instead of a class?**
+
+In v1, only string shorthands and config objects are accepted — no imports needed, zero ceremony. In v2, class instances (`ContextManager`) are also accepted for power users who need lifecycle hooks, direct storage access, or testability — matching the `model` pattern. In v1 the Agent resolves config directly into plugins/tools; in v2 it resolves to a `ContextManager` instance that owns token estimation, budget, and cursors.
+
+**Q: What if models don't effectively use context management tools?**
+
+That's why `"agentic"` is experimental. `"auto"` ships first with proven framework-driven behavior.
+
+**Q: Can users stay on `conversationManager` in v2?**
+
+Yes, with a deprecation warning. It still works — `contextManager` takes precedence if both are set. Removal happens in v3.
+
+**Q: How does Memory relate to context management?**
+
+Memory is a separate primitive (`memoryManager` parameter) with its own mode and storage — out of scope for this doc, covered by Thomas's Memory primitive design. `contextManager` owns L0 ↔ L1 (within-session); `memoryManager` owns L1 → L2 (cross-session knowledge extraction). The Agent brokers the connection — `memoryManager` gets read access to `contextManager`'s L1 for extraction. Users configure each independently.
+
+---
+
+<details>
+<summary><b>Appendix A: Code Examples</b></summary>
+
+### v1 — Auto, minimal
+
+```typescript
+import { Agent } from "@strands-agents/sdk";
+
+const agent = new Agent({
+  contextManager: "auto",
+  tools: [shell, fileRead],
+});
+```
+
+### v1 — Auto, with user's conversation manager
+
+```typescript
+import { Agent } from "@strands-agents/sdk";
+
+const agent = new Agent({
+  contextManager: "auto",
+  conversationManager: new SummarizingConversationManager(),
+  tools: [shell, fileRead],
+});
+```
+
+### v2 — Default (no configuration needed)
+
+```typescript
+import { Agent } from "@strands-agents/sdk";
+
+const agent = new Agent({
+  tools: [shell, fileRead],
+});
+// Context management on by default. SandboxStorage, auto strategy.
+```
+
+### v2 — Custom storage
+
+```typescript
+import { Agent } from "@strands-agents/sdk";
+import { S3Storage } from "@strands-agents/storage-s3";
+
+const agent = new Agent({
+  contextManager: {
+    storage: new S3Storage({ bucket: "my-session" }),
+    strategy: "auto",
+    compression: { threshold: 0.8, protectedMessages: 2 },
+  },
+  tools: [shell, fileRead],
+});
+```
+
+### v2 — Agentic
+
+```typescript
+import { Agent } from "@strands-agents/sdk";
+
+const agent = new Agent({
+  contextManager: "agentic",
+  tools: [shell, fileRead],
+});
+// ContextCompression + ContextNavigation + ContextDelegation active.
+// Agent controls its own L0: compress, pin, search L1 history.
+```
+
+### v2 — Class instance (power user)
+
+```typescript
+import { Agent, ContextManager } from "@strands-agents/sdk";
+import { S3Storage } from "@strands-agents/storage-s3";
+
+const cm = new ContextManager({
+  storage: new S3Storage({ bucket: "my-session" }),
+  strategy: "agentic",
+});
+
+const agent = new Agent({
+  contextManager: cm,
+  tools: [shell, fileRead],
+});
+
+// Direct access to storage (e.g., for Memory integration or testing)
+cm.storage; // S3Storage instance
+await cm.dispose(); // Cleanup when done
+```
+
+### v2 — Opt out
+
+```typescript
+import { Agent } from "@strands-agents/sdk";
+
+const agent = new Agent({
+  contextManager: false,
+  tools: [shell, fileRead],
+});
+```
+
+### v2 — Migration from conversationManager
+
+```typescript
+// Before (v1 — still works with deprecation warning in v2)
+const agent = new Agent({
+  conversationManager: new SummarizingConversationManager(),
+});
+
+// After (v2 — migrated)
+const agent = new Agent({
+  contextManager: { compression: { strategy: "summarize" } },
+});
+```
+
+</details>
+
+---
+
+<details>
+<summary><b>Appendix B: Alternatives Considered</b></summary>
+
+### Class-only primitive (no config object path)
+
+```typescript
+new Agent({
+  contextManager: new ContextManager({
+    storage: new S3Storage(...),
+    strategy: "agentic",
+  }),
+});
+```
+
+**Why rejected as the only path:** Forces an import and instantiation even for the common case. String shorthands (`"auto"`) are simpler. However, class instances ARE accepted as a power-user path — the design supports both (see "Config object or class instance" above).
+
+### Presets as separate Plugin classes
+
+```typescript
+new Agent({ plugins: [new AutoContextManager()] });
+```
+
+**Why rejected:** Lower discoverability. A constructor parameter is easier to find than a plugin import.
+
+### Factory static methods
+
+```typescript
+Agent.withAutoContext({ tools: [...] });
+```
+
+**Why rejected:** Must expose full constructor parameter set. Suggests "different kind of Agent."
+
+### Single unified plugin
+
+```typescript
+new Agent({
+  plugins: [new ContextManager({ mode: "agentic", ... })],
+});
+```
+
+**Why rejected:**
+- God object trajectory as capabilities grow
+- Can't ship plugins independently
+- Preset hides the decomposition anyway
+
+### Single monolithic class (no plugin decomposition)
+
+```typescript
+new Agent({
+  contextManager: new ContextManager({ strategy: "summarize", ... }),
+});
+// Where ContextManager owns compression, caching, navigation, delegation — everything in one class
+```
+
+**Why rejected:**
+- God object — owns everything
+- Fights the composable plugin architecture
+- Limits future additions
+
+Note: We adopted the `ContextManager` name but with a different architecture — it's a thin config resolver that composes separate plugins, not a monolith.
+
+</details>
+
+---
+
+<details>
+<summary><b>Appendix C: Composition Edge Cases</b></summary>
+
+### v1: User provides both preset and conversationManager
+
+```typescript
+new Agent({
+  contextManager: "auto",
+  conversationManager: new SummarizingConversationManager(),
+});
+```
+
+**Behavior:** User's conversation manager is used for compression. `ContextOffloader` still handles tool result caching. No conflict.
+
+### v2: User provides both (deprecated)
+
+```typescript
+new Agent({
+  contextManager: "agentic",
+  conversationManager: new SummarizingConversationManager(),
+});
+```
+
+**Behavior:** `contextManager` takes precedence. Deprecation warning emitted.
+
+### v2: User provides overlapping ContextCompression
+
+```typescript
+new Agent({
+  contextManager: "agentic",
+  plugins: [new ContextCompression({ storage: new S3Storage(...) })],
+});
+```
+
+**Behavior:** User's `ContextCompression` wins (dedup). Preset's `ContextNavigation` still registers.
+
+</details>
+
+---
+
+<details>
+<summary><b>Appendix D: Plugin & Strategy Details</b></summary>
+
+### `ContextManagerConfig`
+
+```typescript
+// On AgentConfig (inline union, no separate type alias — matches model pattern)
+interface AgentConfig {
+  contextManager?: "auto" | "agentic" | ContextManagerConfig | ContextManager | false;
+}
+
+interface ContextManagerConfig {
+  storage?: Storage;                             // default: SandboxStorage
+  strategy?: "auto" | "agentic";                // default: "auto"
+  tokenCountingStrategy?: "auto" | "heuristic"; // default: "auto"
+
+  // ToolResultCache plugin config
+  toolResultCache?: {
+    threshold?: number;                          // token count above which results are cached (TBD)
+    maxAge?: number;                             // auto-eviction time in ms (TBD)
+  } | false;
+
+  // ContextCompression plugin config
+  compression?: {
+    threshold?: number;                          // ratio of context window (0–1]; default: 0.7
+    strategy?: "truncate" | "summarize" | CompressionFn;  // default: "summarize"
+    protectedMessages?: number;                  // default: 1
+  } | false;
+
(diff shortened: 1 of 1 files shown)


designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+Everything `"auto"` does still happens — the agent *also* gets tools to actively manage its own L0. It can decide when to compress, what to protect from eviction, and browse evicted messages in L1.
+
+**Use cases:** Research agents and coding assistants that benefit from self-awareness about their context state. Long-running autonomous agents. Exploratory development where agent autonomy is the point. Parent agents in multi-agent orchestrations.
+
+### Growth and ownership
+
+More strategies may be added as context management capabilities evolve. Per the [pay-for-play tenet](https://github.com/strands-agents/docs/blob/main/team/TENETS.md), the Strands team owns these presets and reserves the right to make breaking changes to them — what a strategy resolves to underneath (which plugins, which defaults) can change between versions. Users who need stability should configure plugins directly rather than relying on preset internals.
+
+Defaults for all preset configurations (thresholds, strategies, token limits) should be informed by benchmarking before shipping. No default should be set based on intuition alone.
```
notowen333, line 87: Definitely agree with the intention here. Do we have ideas for where/when these might run, how to devise the benchmark etc?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+| Tier | What it is | Access cost | Backed by |
+|------|-----------|-------------|-----------|
+| **L0 — Context window** | `agent.messages` — what the model sees on every request. May contain summaries where originals were evicted. | Zero (already in context) | In-memory message list |
+| **L1 — Session history** | Append-only log of messages. Only written when there's a consumer (agentic mode or `memoryManager`). Messages are batch-written to L1 before each summarization. See Appendix E for storage internals. | Low (retrieval tool call) | `ContextManager` storage (`FileStorage`, `S3Storage`, etc.) |
+| **L2 — Long-term memory** | Cross-session knowledge — archival memory, semantic search index, learned facts. Persists beyond the current session. | Higher (query + load from persistent storage) | Memory primitive, vector store, managed memory services |
+
+L2 (long-term memory) is a separate primitive (`memoryManager`) — out of scope for this document. From this doc's perspective, L1 is append-only and immutable for the session lifetime. The tool result cache is separate — short-lived and auto-evicting, not immutable.
```
opieter-aws, line 39: Immutable L1 storage for the session lifetime makes this a non-extensible construct. What is the effect on search time of short-term memory if we have long-running agent sessions? What if the user wants the agent to manage it's short-term memory and re-prioritize/re-score the existing messages in L1?

designs/0010-context-strategy.md
lizradway: Planning on getting some datasets from science team/evals folks and running A/B testing with each strategy and different thresholds.

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+The logic lives in plugins, not tools. Tools are agent-facing wrappers around plugin methods. In auto mode, the framework calls plugin methods via hooks (e.g., `ContextCompression.compress()` fires on threshold). In agentic mode, the same hooks still fire as a safety net, but tools are also registered so the agent can invoke the same methods proactively. The strategy controls which tools are exposed — not what logic exists.
+
+| | `"auto"` | `"agentic"` |
+|--|----------|-------------|
+| Framework compresses when threshold hit | Yes | Yes |
+| Framework caches oversized tool results | Yes | Yes |
+| Agent can recover truncated tool output | Yes (`retrieveToolResult`) | Yes |
+| Agent can trigger compression | No | Yes (`compressContext`) |
+| Agent can protect messages from eviction | No | Yes (`pinMessage`) |
```
zastrowm, line 61: Do we have any numbers on how useful each one of these are?  `pinMessage` itself is interesting which is why I ask

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+See Appendix E for L1 storage internals.
+
+---
+
+## 3. Presets
+
+The `contextManager` parameter accepts a **strategy** that determines who controls L0 — what stays in the context window and when things get compressed to L1.
+
+Both strategies share the same infrastructure. When context pressure builds, L0 is compressed. Oversized tool results are cached separately, and `retrieveToolResult` is always available so the agent can recover truncated outputs without reasoning about context.
```
mkmeral, line 49: > Both strategies share

doc feedback: what both strategies? no intro to strategies here

designs/0010-context-strategy.md
lizradway: Have already done some dumb versions of this, but I think shipping this first class justifies in-depth A/B testing on verified datasets

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+// Custom offloader threshold
+const agent = new Agent({
+  contextManager: { strategy: "auto", contextOffloader: { threshold: 5000 } },
+});
+```
+
+`contextManager` accepts a string shorthand or a config object in v1. The Agent resolves the input directly into plugins and tools — no intermediate class. The `ContextManager` class (owning token estimation, budget, cursors) is introduced in v2 for power users who need lifecycle hooks or direct storage access.
+
+
+In v1, the default is `undefined` (no context management). This avoids surprises on upgrade and gives us time to prove the behavior works before making it default.
```
notowen333, line 120: small detail: I thought we were saying `null` to express the lack of something?

designs/0010-context-strategy.md
opieter-aws, line 49: Will L1 store references to offloaded tools, or the original tool results?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+---
+
+## 3. Presets
+
+The `contextManager` parameter accepts a **strategy** that determines who controls L0 — what stays in the context window and when things get compressed to L1.
+
+Both strategies share the same infrastructure. When context pressure builds, L0 is compressed. Oversized tool results are cached separately, and `retrieveToolResult` is always available so the agent can recover truncated outputs without reasoning about context.
+
+The difference: in **auto**, the agent interacts with content. In **agentic**, the agent reasons about its own context.
```
mkmeral, line 51: why "auto" vs "agentic"? do we have name alternatives?

designs/0010-context-strategy.md
lizradway: > What is the effect on search time of short-term memory if we have long-running agent sessions?

Addressed in appendix E for deletion capacity. 

> What if the user wants the agent to manage it's short-term memory and re-prioritize/re-score the existing messages in L1?

This can be handled in message metadata field. a "priority ranking" in metadata, likely a catch all similar to the hook order priorities.

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+The logic lives in plugins, not tools. Tools are agent-facing wrappers around plugin methods. In auto mode, the framework calls plugin methods via hooks (e.g., `ContextCompression.compress()` fires on threshold). In agentic mode, the same hooks still fire as a safety net, but tools are also registered so the agent can invoke the same methods proactively. The strategy controls which tools are exposed — not what logic exists.
+
+| | `"auto"` | `"agentic"` |
+|--|----------|-------------|
+| Framework compresses when threshold hit | Yes | Yes |
+| Framework caches oversized tool results | Yes | Yes |
+| Agent can recover truncated tool output | Yes (`retrieveToolResult`) | Yes |
+| Agent can trigger compression | No | Yes (`compressContext`) |
```
mkmeral, line 60: can agent evict messages too?

designs/0010-context-strategy.md
opieter-aws, line 51: Does this mean that in `auto`, there is basically no L1, just the existing session manager?

designs/0010-context-strategy.md
lizradway: nope, that is considered a separate 'toolresultcache'. Still accessible to agent via retrieval tool or any file read tools

designs/0010-context-strategy.md
opieter-aws: I see this is answered below

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+| Agent can read from L1 | No | Yes (`getHistory`) |
+| Agent can search L1 | No | Yes (`searchHistory`) |
+| Agent can spawn context-aware children | No | Yes (`delegateWithContext`) |
+| Agent can check its context budget | No | Yes (`getContextBudget`) |
+
+Users who don't want presets can build custom context management using the same infrastructure. Token tracking, message metadata, token estimation, and context limit primitives together give users everything they need to write their own `BeforeModelCallEvent` hooks with custom policies. The presets are opinionated compositions of these primitives, not the only way to use them.
+
+### `"auto"` — The framework decides
+
+The framework manages L0 transparently. When context pressure builds, messages are summarized or dropped in L0. No L1 writes happen in auto mode — L1 is only written when there's a consumer (agentic mode or `memoryManager`). The agent doesn't know context management is happening — it just never hits a wall. The only tool exposed is `retrieveToolResult` for recovering truncated tool outputs.
```
zastrowm, line 71: Why not; if we give it access to retrieveToolResult did we also consider giving it access to retrieveOldMessages or something?

designs/0010-context-strategy.md
lizradway: i didn't look into specific alternatives because to me these were really simple and made sense. open to suggestions if you don't like them

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+| **L1 — Session history** | Append-only log of messages. Only written when there's a consumer (agentic mode or `memoryManager`). Messages are batch-written to L1 before each summarization. See Appendix E for storage internals. | Low (retrieval tool call) | `ContextManager` storage (`FileStorage`, `S3Storage`, etc.) |
+| **L2 — Long-term memory** | Cross-session knowledge — archival memory, semantic search index, learned facts. Persists beyond the current session. | Higher (query + load from persistent storage) | Memory primitive, vector store, managed memory services |
+
+L2 (long-term memory) is a separate primitive (`memoryManager`) — out of scope for this document. From this doc's perspective, L1 is append-only and immutable for the session lifetime. The tool result cache is separate — short-lived and auto-evicting, not immutable.
+
+See Appendix E for L1 storage internals.
+
+---
+
+## 3. Presets
```
mkmeral, line 45: What's the devx here? what are the alternatives instead of these 2 presets? What are the pros/cons of each?

designs/0010-context-strategy.md
zastrowm, line 71: With auto-mode (auto-summarization) we should also consider observability - e.g. how are consumers of the agent aware that this happenend - both in terms of OTEL but also just wanting to do something at the application layer in response.  Hooks would probably be a way to surface the latter

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+> **Note:** `"agentic"` is experimental. It depends on research into whether models effectively use context management tools. `"auto"` ships first and is the primary focus.
+
+Everything `"auto"` does still happens — the agent *also* gets tools to actively manage its own L0. It can decide when to compress, what to protect from eviction, and browse evicted messages in L1.
+
+**Use cases:** Research agents and coding assistants that benefit from self-awareness about their context state. Long-running autonomous agents. Exploratory development where agent autonomy is the point. Parent agents in multi-agent orchestrations.
+
+### Growth and ownership
+
+More strategies may be added as context management capabilities evolve. Per the [pay-for-play tenet](https://github.com/strands-agents/docs/blob/main/team/TENETS.md), the Strands team owns these presets and reserves the right to make breaking changes to them — what a strategy resolves to underneath (which plugins, which defaults) can change between versions. Users who need stability should configure plugins directly rather than relying on preset internals.
```
mkmeral, line 85: > the Strands team owns these presets and reserves the right to make breaking changes to them

whoa whoa whoa. that's a big statement to make. do we actually want to make it yet? will auto change? can auto become agentic later on if the framework decides?

designs/0010-context-strategy.md
lizradway: TK in https://github.com/strands-agents/sdk-python/issues/2225

only have run rudimentary agent scripts, but full benchmarking tracked above

designs/0010-context-strategy.md
lizradway: works for me! can update this

designs/0010-context-strategy.md
notowen333, line 71: Should we also expose something like `batchRetrieveToolResults` or otherwise support batching in the proposed tool?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+## 4. TypeScript v1
+
+Ships `contextManager: "auto"` as an opt-in parameter.
+
+### What ships
+
+```typescript
+// Minimal — defaults to SandboxStorage
+const agent = new Agent({ contextManager: "auto" });
```
poshinchen, line 99: `{ contextManager: "auto" }` I'm fine with the `strategy: "auto` below, but not this one. It looks misleading.

designs/0010-context-strategy.md
zastrowm, line 85: >  Per the [pay-for-play tenet] 

We neither have this tenet in the linked doc nor doe I think that this is pay-for-play in terms of "the Strands team owns these presets and reserves the right to make breaking changes to them "

There's probably a decision/something that needs to be added around opinionated defaults, but I'm not sure this is a case where we can just say "reserves the right to make breaking changes to them".  A discussion point perhaps

designs/0010-context-strategy.md
lizradway: see appendix B and appendix A

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+  contextManager: { strategy: "auto", contextOffloader: false },
+});
+
+// Custom offloader threshold
+const agent = new Agent({
+  contextManager: { strategy: "auto", contextOffloader: { threshold: 5000 } },
+});
+```
+
+`contextManager` accepts a string shorthand or a config object in v1. The Agent resolves the input directly into plugins and tools — no intermediate class. The `ContextManager` class (owning token estimation, budget, cursors) is introduced in v2 for power users who need lifecycle hooks or direct storage access.
```
mkmeral, line 117: why separate work for v1 and v2? should we just do v2?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+The framework manages L0 transparently. When context pressure builds, messages are summarized or dropped in L0. No L1 writes happen in auto mode — L1 is only written when there's a consumer (agentic mode or `memoryManager`). The agent doesn't know context management is happening — it just never hits a wall. The only tool exposed is `retrieveToolResult` for recovering truncated tool outputs.
+
+**Use cases:** Beginners who don't want to think about context yet. Production deployments where predictability matters more than autonomy. Cost-sensitive workloads at scale. Multi-agent systems where child agents need lightweight, hands-off management. Anyone who wants it to just not break.
+
+### `"agentic"` — The agent decides
+
+> **Note:** `"agentic"` is experimental. It depends on research into whether models effectively use context management tools. `"auto"` ships first and is the primary focus.
+
+Everything `"auto"` does still happens — the agent *also* gets tools to actively manage its own L0. It can decide when to compress, what to protect from eviction, and browse evicted messages in L1.
```
opieter-aws, line 79: In this state, does the framework still compress when threshold hits? Or does the agent then decide what happens? (I would expect the latter with an agentic mode)

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+**Use cases:** Research agents and coding assistants that benefit from self-awareness about their context state. Long-running autonomous agents. Exploratory development where agent autonomy is the point. Parent agents in multi-agent orchestrations.
```
opieter-aws, line 81: We should consider giving the agent full autonomy over L1 for an agentic mode

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+| Component | What it does |
+|-----------|-------------|
+| **`ContextOffloader`** | Oversized tool results cached separately (short-lived, auto-evicting) via `AfterToolCallEvent` hook. Agent sees a truncated preview. Provides `retrieveToolResult` tool for on-demand access. Renamed to `ToolResultCache` in v2 (config key changes from `contextOffloader` to `toolResultCache`). |
+| **Proactive compression on conversation manager** | `BeforeModelCallEvent` hook checks L0 token usage against threshold. When exceeded, compresses older messages in L0. |
+| **Message protection** | Position-based pinning (`protectedMessages = 1` by default) ensures the task prompt is never evicted. |
+
+### Coexistence with `conversationManager`
+
+In v1, the user's conversation manager is respected. The preset configures proactive compression *on* whichever conversation manager the user provides (or the default):
```
mkmeral, line 132: what about v2? how do they relate? which one should we tell customers to use, and why? does it cover all conversation manager use cases?

designs/0010-context-strategy.md
poshinchen: "enhanced"?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+```
+
+---
+
+## 5. TypeScript v2
+
+Three changes:
+1. **`"auto"` becomes the default** — every agent gets context management unless opted out
+2. **`"agentic"` ships** — agent-in-the-loop context management
+3. **`conversationManager` is deprecated** — its compression responsibility moves to `ContextCompression`
```
zastrowm, line 152: Why deprecated?  Can we just remove it given that we have proper paths forward?

designs/0010-context-strategy.md
lizradway: see appendix e for a full explanation of the "sliding window index" system proposed for messages tracking. deletion can be configured for long running agents.

but tldr; basically no, agent cannot independently decide to evict messages

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+In v1, only string shorthands and config objects are accepted — no imports needed, zero ceremony. In v2, class instances (`ContextManager`) are also accepted for power users who need lifecycle hooks, direct storage access, or testability — matching the `model` pattern. In v1 the Agent resolves config directly into plugins/tools; in v2 it resolves to a `ContextManager` instance that owns token estimation, budget, and cursors.
+
+**Q: What if models don't effectively use context management tools?**
+
+That's why `"agentic"` is experimental. `"auto"` ships first with proven framework-driven behavior.
+
+**Q: Can users stay on `conversationManager` in v2?**
+
+Yes, with a deprecation warning. It still works — `contextManager` takes precedence if both are set. Removal happens in v3.
```
opieter-aws, line 203: Why not remove in V2?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+});
+```
+
+---
+
+## 5. TypeScript v2
+
+Three changes:
+1. **`"auto"` becomes the default** — every agent gets context management unless opted out
+2. **`"agentic"` ships** — agent-in-the-loop context management
```
mkmeral, line 151: wait so we didn't have agentic in v1? so the only thing we'd have is "auto"? so auto and disabled are the two options?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+Users who don't want presets can build custom context management using the same infrastructure. Token tracking, message metadata, token estimation, and context limit primitives together give users everything they need to write their own `BeforeModelCallEvent` hooks with custom policies. The presets are opinionated compositions of these primitives, not the only way to use them.
+
+### `"auto"` — The framework decides
+
+The framework manages L0 transparently. When context pressure builds, messages are summarized or dropped in L0. No L1 writes happen in auto mode — L1 is only written when there's a consumer (agentic mode or `memoryManager`). The agent doesn't know context management is happening — it just never hits a wall. The only tool exposed is `retrieveToolResult` for recovering truncated tool outputs.
+
+**Use cases:** Beginners who don't want to think about context yet. Production deployments where predictability matters more than autonomy. Cost-sensitive workloads at scale. Multi-agent systems where child agents need lightweight, hands-off management. Anyone who wants it to just not break.
+
+### `"agentic"` — The agent decides
```
yonib05, line 75: With agentic, do we want to support the passing in of an agent class?

Example: For the compaction I want to use Haiku with a specific prompt or context that is focused on outputting something optimized for my use case vs. just giving the orchestrator model a tool.

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+Two changes: (1) `"auto"` becomes the default — agents that didn't set `contextManager` now get it automatically. Users who don't want this can set `contextManager: false`. (2) Internals upgrade — `ToolResultCache` replaces `ContextOffloader`, `ContextCompression` replaces conversation manager compression — same external behavior.
+
+**Q: What if the user provides their own plugin?**
+
+Deduplication detects by type and skips the preset's version. User's `ToolResultCache`, `ContextCompression`, or `ContextNavigation` (v2) wins.
+
+**Q: Why a config object instead of a class?**
+
+In v1, only string shorthands and config objects are accepted — no imports needed, zero ceremony. In v2, class instances (`ContextManager`) are also accepted for power users who need lifecycle hooks, direct storage access, or testability — matching the `model` pattern. In v1 the Agent resolves config directly into plugins/tools; in v2 it resolves to a `ContextManager` instance that owns token estimation, budget, and cursors.
```
mkmeral, line 195: why not just do v2 directly? why do we spend time with v1?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+v2 introduces `OnContextOverflowEvent` — a new lifecycle event fired on context length errors, replacing the opaque retry logic in conversation managers. The SDK already normalizes these errors across all providers into `ContextWindowOverflowError` — this event just exposes that to plugins.
+
+### `conversationManager` deprecation
+
+In v2, `ContextCompression` takes over compression and L1 writing. Nothing left for `conversationManager`.
+
+**Deprecation path:**
+1. **v1 (coexist):** Both live side by side. No conflict.
+2. **v2 (deprecate):** Deprecation warning emitted. `contextManager` takes precedence if both are set.
```
yonib05, line 174: I would accelerate this and remove it in v2 and deprecate it in the minor versions leading to v2.

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+## 6. Anticipated Questions
+
+**Q: Why isn't `"auto"` the default in v1?**
+
+We need to prove the behavior first. v1 is opt-in so we can validate with early adopters. v2 makes it the default once we're confident it works reliably.
+
+**Q: What happens when users upgrade from v1 to v2?**
+
+Two changes: (1) `"auto"` becomes the default — agents that didn't set `contextManager` now get it automatically. Users who don't want this can set `contextManager: false`. (2) Internals upgrade — `ToolResultCache` replaces `ContextOffloader`, `ContextCompression` replaces conversation manager compression — same external behavior.
```
zastrowm, line 187: nit: `"contextManager: 'none'"` would make some sense given our stringly-ness already.  It's a predefined "preset" that just doesn't do anything

designs/0010-context-strategy.md
mkmeral, line 203: wait why? why do we have duplicated concepts? what are the pros and cons?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+|--|----------|-------------|
+| Framework compresses when threshold hit | Yes | Yes |
+| Framework caches oversized tool results | Yes | Yes |
+| Agent can recover truncated tool output | Yes (`retrieveToolResult`) | Yes |
+| Agent can trigger compression | No | Yes (`compressContext`) |
+| Agent can protect messages from eviction | No | Yes (`pinMessage`) |
+| Agent can read from L1 | No | Yes (`getHistory`) |
+| Agent can search L1 | No | Yes (`searchHistory`) |
+| Agent can spawn context-aware children | No | Yes (`delegateWithContext`) |
+| Agent can check its context budget | No | Yes (`getContextBudget`) |
```
gautamsirdeshmukh, line 65: Interested to know how this `getContextBudget` tool has held up in practice. My mind specifically goes to 2 concerns:
1. "Context fright" (when agents are aware that they are approaching a limit, they can rush / reason less / produce lower quality output)
2. The agent checking its context usage between / during turns far more than is necessary

Additionally, if the agent is able to decide when to compress / what messages to "pin", does this mean the agentic management strategy would be used separately from proactive compression? I'm thinking about cases where a proactive compression operation is underway but the agent is unaware of it and decides to manage compression itself..

designs/0010-context-strategy.md
notowen333, line 75: Is this part of existing STOA approaches or it's more experimental?

designs/0010-context-strategy.md
zastrowm, line 195: Why class only in v2?

designs/0010-context-strategy.md
lizradway: that is essentially the "agentic" approach. i thought about this, but i think the difference is that `retrieveToolResult` retrieves a tool result hat it _should_ have full access to anyways. whereas querying message history is an agentic capability/offloading old messages is standard + expected practice.

happy to hear argument to include `retrieveOldMessages` in auto and would be open to considering, since i went back and forth alot

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+  conversationManager: new SummarizingConversationManager(),
+});
+```
+
+---
+
+## 5. TypeScript v2
+
+Three changes:
+1. **`"auto"` becomes the default** — every agent gets context management unless opted out
```
yonib05, line 150: Why not agentic as the default if we want to push and challenge the leading edge?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+**Q: What if the user provides their own plugin?**
+
+Deduplication detects by type and skips the preset's version. User's `ToolResultCache`, `ContextCompression`, or `ContextNavigation` (v2) wins.
+
+**Q: Why a config object instead of a class?**
+
+In v1, only string shorthands and config objects are accepted — no imports needed, zero ceremony. In v2, class instances (`ContextManager`) are also accepted for power users who need lifecycle hooks, direct storage access, or testability — matching the `model` pattern. In v1 the Agent resolves config directly into plugins/tools; in v2 it resolves to a `ContextManager` instance that owns token estimation, budget, and cursors.
+
+**Q: What if models don't effectively use context management tools?**
```
zastrowm, line 197: Yeah this one is interesting; also what if some do well and some suck; I think the default proposed makes a lot more sense

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+| Component | Role | Domain | Tools | Mode |
+|-----------|------|--------|-------|------|
+| **`ContextManager`** | Config resolver + shared infrastructure | Storage, token estimation, budget, cursors | `getContextBudget` (agentic) | Always |
+| **`ToolResultCache`** | Plugin | Cache oversized tool results | `retrieveToolResult` | Both |
+| **`ContextCompression`** | Plugin | L0 writes — compression, pinning | `compressContext`, `pinMessage` (agentic) | Both |
+| **`ContextNavigation`** | Plugin | Read from L1 (session history) | `getHistory`, `searchHistory` | Agentic only |
+| **`ContextDelegation`** | Plugin | Context-aware child spawning | `delegateWithContext` | Agentic only |
+
+v2 introduces `OnContextOverflowEvent` — a new lifecycle event fired on context length errors, replacing the opaque retry logic in conversation managers. The SDK already normalizes these errors across all providers into `ContextWindowOverflowError` — this event just exposes that to plugins.
```
opieter-aws, line 166: Can we also introduce an `onEviction` event, for when messages move from L0 to L1?

designs/0010-context-strategy.md
pgrayy, line 65: Minor consideration: Do we want a suite of tools or a single tool with different modes. From an organizational stand point, I think it might be cleaner to have a single "context" tool that could have different modes like "search", "pin", "compress", etc.

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+```
+
+</details>
+
+---
+
+<details>
+<summary><b>Appendix B: Alternatives Considered</b></summary>
+
+### Class-only primitive (no config object path)
```
mkmeral, line 330: what does class only mean? what is in that class? how do i use that class?

designs/0010-context-strategy.md
opieter-aws: Does this mean that the metadata field is mutable?

designs/0010-context-strategy.md
lizradway: is this raised currently by surfacing the context overflow error? 

will think about this more, since im not sure what the ideal devx would be

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+**Why rejected as the only path:** Forces an import and instantiation even for the common case. String shorthands (`"auto"`) are simpler. However, class instances ARE accepted as a power-user path — the design supports both (see "Config object or class instance" above).
+
+### Presets as separate Plugin classes
+
+```typescript
+new Agent({ plugins: [new AutoContextManager()] });
+```
+
+**Why rejected:** Lower discoverability. A constructor parameter is easier to find than a plugin import.
```
mkmeral, line 349: why though? if we double down on plugins like OpenAIs capabilities, can't we solve this issue? should we do it?

designs/0010-context-strategy.md
lizradway: what is the use case for this? retrieve tool results only _should_ typically be happening immediately after tool truncation if the agent needs more than the preview

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+### v2 — Custom storage
+
+```typescript
+import { Agent } from "@strands-agents/sdk";
+import { S3Storage } from "@strands-agents/storage-s3";
+
+const agent = new Agent({
+  contextManager: {
+    storage: new S3Storage({ bucket: "my-session" }),
```
zastrowm, line 256: This begs the question what storage does here.

This seems more like a memory concern?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+### Single unified plugin
+
+```typescript
+new Agent({
+  plugins: [new ContextManager({ mode: "agentic", ... })],
+});
+```
+
+**Why rejected:**
+- God object trajectory as capabilities grow
```
mkmeral, line 368: why? is that a problem? we can split later 

> Can't ship plugins independently

wdym?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+## 5. TypeScript v2
+
+Three changes:
+1. **`"auto"` becomes the default** — every agent gets context management unless opted out
+2. **`"agentic"` ships** — agent-in-the-loop context management
+3. **`conversationManager` is deprecated** — its compression responsibility moves to `ContextCompression`
+
+### Plugin decomposition
+
+Each plugin has a single cohesive responsibility. `ContextManager` is not a plugin — it's the top-level class that the config resolves to. It owns shared infrastructure (storage, token estimation, budget tracking, cursors) and composes the plugins below. Plugins read from `ContextManager` rather than depending on each other.
```
yonib05, line 156: What plugins would we vend?

AgentCore Memory, we would partner with the AgentCore SDK for, but I still feel like we should vend some more strategies to show the breadth of the capabilities in strands.

designs/0010-context-strategy.md
lizradway: >  can auto become agentic later on if the framework decides?

no i think we should define general expectations, but defaults can be updated if the user is opted in (unlike low level defaults which i view as breaking)

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+  contextManager: new ContextManager({
+    storage: new S3Storage(...),
+    strategy: "agentic",
+  }),
+});
+```
+
+**Why rejected as the only path:** Forces an import and instantiation even for the common case. String shorthands (`"auto"`) are simpler. However, class instances ARE accepted as a power-user path — the design supports both (see "Config object or class instance" above).
+
+### Presets as separate Plugin classes
```
zastrowm, line 343: We should also consider the "Can you have multiples of these" as an answer to this question.  I presume we wouldn't want two ContextManagers, so it makes sense as a parameter for that reason

designs/0010-context-strategy.md
lizradway: my personal opinion is that low level default changes are breaking, high level default changes should be expected.

but i think this is a good discussion point

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+  contextManager: { compression: { strategy: "summarize" } },
+});
+```
+
+</details>
+
+---
+
+<details>
+<summary><b>Appendix B: Alternatives Considered</b></summary>
```
mkmeral, line 328: doc feedback: I see why you rejected certain ideas, but that does not tell me what those ideas are and if you rejecting them is valid. I need more on these options. I personally am not 100% bought into the proposal, and I cannot invalidate the other approaches because they are not explained

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+## 2. The Cache Hierarchy
+
+> **Note:** This doc covers L0 ↔ L1 (within-session context management). L2 and anything memory-related is out of scope — see Thomas's Memory primitive design. References to memory in this doc are included only for background.
+
+Context management maps to a three-tier cache hierarchy. Every operation is movement between tiers.
+
+| Tier | What it is | Access cost | Backed by |
+|------|-----------|-------------|-----------|
+| **L0 — Context window** | `agent.messages` — what the model sees on every request. May contain summaries where originals were evicted. | Zero (already in context) | In-memory message list |
+| **L1 — Session history** | Append-only log of messages. Only written when there's a consumer (agentic mode or `memoryManager`). Messages are batch-written to L1 before each summarization. See Appendix E for storage internals. | Low (retrieval tool call) | `ContextManager` storage (`FileStorage`, `S3Storage`, etc.) |
```
pgrayy, line 36: Is there more than just messages we consider "context" and thus would want to manage?

pgrayy, line 36: Are you tying this into the transcript construct @JackYPCOnline proposed?

designs/0010-context-strategy.md
pgrayy, line 75: Has this been tested/prototyped yet? How exactly does the agent decide to context manage? What does it look at and how often is it looking at it? Would the agent be wasting tokens frequently checking its context when not necessary?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+---
+
+<details>
+<summary><b>Appendix E: L1 Storage Model</b></summary>
+
+> **Note:** This is an implementation detail and subject to change. The external behavior (append-only L1, cursor-scoped reads) is stable; the physical storage layout is not.
+
+### Physical storage
+
+When L1 has a consumer (agentic mode or `memoryManager`), messages are batch-written to L1 before each summarization. Each batch is a single file containing the full `messages` array at that point. Each message carries a **turn ID** in metadata — a monotonically increasing identifier assigned per agent turn. In auto mode without `memoryManager`, no L1 writes occur.
```
opieter-aws, line 554: How would L1 know that it has a consumer in case of `memoryManager`, which handles extraction independently?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+
+```
+storage/
+  batch-001.json   # messages from turns 1–25 (written before first summarization)
+  batch-002.json   # messages from turns 26–50 (written before second summarization)
+  ...
```
opieter-aws, line 560: How searchable is this structure for the agent?

designs/0010-context-strategy.md
```diff
@@ -0,0 +1,602 @@
...
+```
+
+Messages are written once. Each batch contains only new messages since the previous batch.
+
+### Cursors
+
+A **cursor** is a start pointer into the flat log — a turn ID. Everything from that turn forward is the agent's visible L1.
```
opieter-aws, line 567: Can the agent decide to set cursors that are not purely positional?

designs/0010-context-strategy.md
lizradway: i was planning on adding message metadata as an extensible optional field (for priority, etc). offloaded large tool results are available already using grep/`retrieveToolResult`

designs/0010-context-strategy.md
lizradway: i think so

designs/0010-context-strategy.md
lizradway: conclusion was to have many more strategies, so more specific naming is preferred ie. `agenticContextTools`.

designs/0010-context-strategy.md
lizradway: > "Context fright" (when agents are aware that they are approaching a limit, they can rush / reason less / produce lower quality output)

Fair, but im not sure if this is bad or just a trade off of quality for context (which is all context management is).

> The agent checking its context usage between / during turns far more than is necessary

Vending `current context / context window limit` in system prompt should ease this.

>Additionally, if the agent is able to decide when to compress / what messages to "pin", does this mean the agentic management strategy would be used separately from proactive compression? 

nah, agent can decide when to compress via tool wrapper around compression plugin. i think due to the step wise operation of our agentic loop, this isnt possible: "proactive compression operation is underway but the agent is unaware of it and decides to manage compression itself"

lizradway: i really really really like the idea of a toolbox/groups of tools that can be vended as we discussed. 


im not quite sure what this would look like concretely, but am a big endorser philosophically. will try to follow this in v2

designs/0010-context-strategy.md
lizradway: what does STOA stand for :,)

lizradway: i think this would make a lot of sense for a user to be able to pass this as a custom `CompressionConfig` strategy. users could not only then compress using whatever prompt/model/etc they want, but as long as it implements `reduce` (and optionally `store` for L1) they can use whatever context compression method (other than just sliding window and summarizing) they chose with Strands.

designs/0010-context-strategy.md
lizradway: interesting. i was planning on the former, but you have a good point with the latter. i think this could be informed by benchmarks for a final decision.

designs/0010-context-strategy.md
lizradway: new `agenticContextModification` strategy perhaps ?

designs/0010-context-strategy.md
lizradway: easy win to add the shorthand in v1 to gain visibility and get some initial signal into user default preferences since we are setting this for them.

designs/0010-context-strategy.md
lizradway: In v2, we should be telling customers to use the contextManager and will be issuing a deprecation warning for the conversation manager.

v1 there are no changes

designs/0010-context-strategy.md
lizradway: i think we can market the frontier use cases more than auto, but it goes against standard and i would argue would be a security nightmare for us to default `--dangerously-allow-XXXX` to true.  The user needs to explicitly opt into risk.

designs/0010-context-strategy.md
lizradway: yes

designs/0010-context-strategy.md
lizradway: fine by me. would we want to issue deprecation warning now then? or is this fine to cut off in MV upgrade?

designs/0010-context-strategy.md
lizradway: AgentCore Memory is more in Thomas's memory purview, so I would recommend speaking to him about that.


full plugin list should be listed below.

designs/0010-context-strategy.md
lizradway: followed up offline and yes love it

designs/0010-context-strategy.md
lizradway: sounds good -- seems like mackenzie is in agreement as well. will release deprecation warning in late v1 to remove fully in v2.

designs/0010-context-strategy.md
lizradway: answered in https://github.com/strands-agents/docs/pull/831/changes#r3221511006

lizradway: would require breaking changes to move token estimation logic into the class. not necessary for auto mode anyways imo

designs/0010-context-strategy.md
lizradway: concluded were all aligned to remove in v2, deprecation warning in late v1

lizradway: concluded were all aligned to remove in v2, deprecation warning in late v1

designs/0010-context-strategy.md
lizradway: L1 storage ie. where tool results are offloaded to and messages are written out pre-summarization

designs/0010-context-strategy.md
lizradway: exposing class on public api rather than config

designs/0010-context-strategy.md
lizradway: makes sense

designs/0010-context-strategy.md
lizradway: plugins array will start to scale to absolutely unmaintainable heights
no priority or separation of concerns, just one big bag of plugins, etc

designs/0010-context-strategy.md
lizradway: we cannot ship only context offloader or only compression or only proactive compression independently. users MUST define this in the ContextManager rather than independently.

PR #839, opened 2026-05-14
feat(offloader): add search results tool

lizradway:
## Description

Documents the new search and filtering capabilities added to the `retrieve_offloaded_content` tool in the TypeScript SDK ([strands-agents/sdk-typescript#1060](https://github.com/strands-agents/sdk-typescript/pull/1060)).

The retrieval tool now supports optional `pattern`, `line_range`, and `context_lines` parameters for targeted retrieval instead of fetching the entire offloaded blob.

Changes:
- Promoted "Retrieval Tool" from a subsection under "How It Works" to its own top-level `##` section
- Added TypeScript-specific documentation of the new tool parameters and retrieval modes
- Added a step-by-step example showing what the agent sees (offloaded result, pattern search, line range access, truncation)
- Added "Disabling the retrieval tool" subsection

## Related Issues

https://github.com/strands-agents/sdk-typescript/pull/1060

## Type of Change

- [x] Content update/revision
- [x] Structure/organization improvement

## Checklist

----

### DIFF
--- a/src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
+++ b/src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
@@ -333,7 +333,7 @@ Input: `{ reference: "mem_1_tool-123_0", pattern: "admin", context_lines: 2 }`
 
 ### Using other tools for retrieval
 
-When using `FileStorage`, the agent can use its existing tools (shell, grep, cat, etc.) to access offloaded content directly from the file system:
+When using `FileStorage`, the agent can use its existing tools (shell, grep, cat, etc.) to access offloaded content directly from the file system. The offloaded guidance includes the full storage path, so the agent knows where to look:
 
 ```
 grep -n "admin" ./artifacts/mem_1_tool-123_0
@@ -356,12 +356,17 @@ This approach is often preferable because the agent already knows these tools we
 <Tab label="Python">
 
 ```python
-agent = Agent(plugins=[
-    ContextOffloader(
-        storage=FileStorage("./artifacts"),
-        include_retrieval_tool=False,
-    )
-])
+from strands_tools import shell
+
+agent = Agent(
+    tools=[shell],
+    plugins=[
+        ContextOffloader(
+            storage=FileStorage("./artifacts"),
+            include_retrieval_tool=False,
+        )
+    ]
+)
 ```
 </Tab>
 <Tab label="TypeScript">
--- a/src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
+++ b/src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
@@ -233,8 +233,6 @@ The plugin includes a `retrieve_offloaded_content` tool that lets the agent fetc
 
 The inline guidance in offloaded results tells the agent to use its available tools to selectively access the data it needs, and mentions `retrieve_offloaded_content` as a fallback.
 
-In Python, the retrieval tool returns the full offloaded content. Search and filtering parameters are not yet available — use the agent's other tools (file readers, shell, etc.) for targeted access.
-
 </Tab>
 <Tab label="TypeScript">
 
@@ -287,6 +285,10 @@ Input: `{ reference: "mem_1_tool-123_0" }`
 
 The tool returns the full offloaded content in its native format.
 
+```
+{"users":[{"id":1,"name":"Alice","role":"admin"},{"id":2,"name":"Bob","role":"user"},{"id":3,"name":"Charlie","role":"user"}, ...]}
+```
+
 </Tab>
 <Tab label="TypeScript">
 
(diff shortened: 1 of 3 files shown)


src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
```diff
@@ -253,6 +224,171 @@ agent = Agent(plugins=[
...
+## Retrieval Tool
+
+The plugin includes a `retrieve_offloaded_content` tool that lets the agent fetch offloaded content by reference, returning it in its native format — text as a string, JSON as a JSON block, images as image blocks, and documents as document blocks. This tool is registered by default.
+
+<Tabs>
+<Tab label="Python">
+
+The inline guidance in offloaded results tells the agent to use its available tools to selectively access the data it needs, and mentions `retrieve_offloaded_content` as a fallback.
+
+In Python, the retrieval tool returns the full offloaded content. Search and filtering parameters are not yet available — use the agent's other tools (file readers, shell, etc.) for targeted access.
```
mkmeral: > Search and filtering parameters are not yet available — use the agent's other tools (file readers, shell, etc.) for targeted access.

we don't need this in Python tab imho

src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
```diff
@@ -253,6 +224,171 @@ agent = Agent(plugins=[
...
+```
+
+**3. Agent reads a specific line range**
+
+Input: `{ reference: "mem_1_tool-123_0", line_range: { start: 45, end: 55 } }`
+
+```
+[Lines 45-55 of 200]
+
+  45|     { "id": 14, "name": "Mallory", "role": "user" },
```
mkmeral: do we need this many examples in detail? I don't think you need to explain how regex search and line ranges work lol

src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
```diff
@@ -253,6 +224,139 @@ agent = Agent(plugins=[
...
+  48|     { "id": 15, "name": "Dana", "role": "user" },
+> 49|     { "id": 16, "name": "Eve", "role": "admin" },
+  50|     { "id": 17, "name": "Frank", "role": "user" }
+  51|   ]
+```
+
+</Tab>
+</Tabs>
+
+### Disabling the retrieval tool
```
mkmeral: nit: do we need a new section for this? also for the "what agent sees"? or is it in retrieval tool section?

src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
```diff
@@ -253,6 +224,139 @@ agent = Agent(plugins=[
...
+The tool returns the full offloaded content in its native format.
+
+```
+{"users":[{"id":1,"name":"Alice","role":"admin"},{"id":2,"name":"Bob","role":"user"},{"id":3,"name":"Charlie","role":"user"}, ...]}
+```
+
+</Tab>
+<Tab label="TypeScript">
+
+**1. Tool result gets offloaded (replaces original result inline)**
```
mkmeral, line 296: while reading https://d3ehv1nix5p99z.cloudfront.net/pr-cms-839/docs/user-guide/concepts/plugins/context-offloader/#retrieval-tool 

I feel like this section is a bit distracting. Do people want to know how it works, or just how to use this? Can we add this to somewhere else?

mkmeral (commented):
Image

Overall the better approach is to give your agent investigation tools (like shell, grep ,etc) for context offloading. we don't talk about that at all, instead we spend half of the page explaining details of a sub-tool. 

I think we should focus more on telling users to give their agents tools to investigate (like skills do for reading/executing), and we should quickly explain that this is what they can use instead, if they want a limited tool to get offloaded content

src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
lizradway: reframed this to be about built in tool use instead

src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
lizradway: I think it is valuable to include in documentation what the tool does/what the results look like so that this is not an opaque black box

src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
```diff
@@ -253,6 +225,161 @@ agent = Agent(plugins=[
...
+With `S3Storage`, the agent can use the AWS CLI to access offloaded content:
+
+```
+aws s3 cp s3://my-agent-artifacts/tool-results/mem_1_tool-123_0 - | grep -n "admin"
+aws s3 cp s3://my-agent-artifacts/tool-results/mem_1_tool-123_0 - | head -50
+```
+
+With `InMemoryStorage`, there is no external access path — the built-in retrieval tool is the only way to access offloaded content, so keep it enabled.
+
+This approach is often preferable because the agent already knows these tools well and can chain them together for complex queries. To disable the built-in retrieval tool and rely on the agent's own tools:
```
mkmeral, line 354: nit: if it's preferable, i'd expect it higher up

mkmeral (approved):

src/content/docs/user-guide/concepts/plugins/context-offloader.mdx
lizradway: fair. im about to release context manager class and these docs are getting restructures then so can put more emphasis / move up in that.

PR #840, opened 2026-05-14
feat: add tags footer at the end of user guide pages

notowen333:
## Description

Footer on every user guide page with tags pills. Each pill is clickable and shows all of the related pages.

## Related Issues

n/a

## Type of Change

- Other (please describe):

## Checklist

----

### DIFF
--- a/src/components/PageTags.astro
+++ b/src/components/PageTags.astro
@@ -2,13 +2,11 @@
 /**
  * Clickable tag pills at the bottom of user-guide articles.
  *
- * Pill row stays static — never reflows. Clicking a pill opens a single
- * shared panel below the row showing every other user-guide page carrying
- * that tag. Click the same pill again (or the panel's close button) to
- * collapse.
- *
- * State is driven by a small script that toggles an `is-active` class on
- * the clicked pill and the `hidden` attribute on each panel.
+ * The pill row is the only thing that ever takes layout space. Clicking a
+ * pill reveals a single shared panel that floats *over* the page below the
+ * row (absolutely positioned), so opening or closing it doesn't push the
+ * footer up or down. Click the pill again, the close button, the ESC key,
+ * or anywhere outside the panel to dismiss.
  *
  * Uses an `aria-expanded` disclosure pattern (each pill is a disclosure
  * button revealing its panel). Not a full ARIA tablist — that would
@@ -16,6 +14,7 @@
  * we don't currently implement.
  */
 import { getCollection } from 'astro:content'
+import { slug as githubSlug } from 'github-slugger'
 import { userGuidePagesWithTag } from '../util/related-docs'
 import { pathWithBase } from '../util/links'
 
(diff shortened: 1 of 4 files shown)


src/components/overrides/MarkdownContent.astro
zastrowm, line 1: I like the UX!

src/components/overrides/MarkdownContent.astro
```diff
@@ -15,4 +16,5 @@ const { languages, community, experimental } = starlightRoute.entry.data;
       contextualizes the page, and language support is listed in the community catalog table instead. */}
   {languages && !community && <LanguageSupportAside languages={languages} />}
   <slot />
+  <PageTags />
```
zastrowm, line 19: Did you play around with putting this at the top at all (or sidebar); I'm curious if that helps people jump to related things or understand what the topic is about better or if it's just noise?

src/components/PageTags.astro
```diff
@@ -0,0 +1,265 @@
...
+        ) : (
+          <p class="page-tags__empty">No other pages with this tag.</p>
+        )}
+      </div>
+    ))}
+  </aside>
+)}
+
+<script>
+  function init() {
```
zastrowm, line 83: Did try with a popover instead by chance?  The "click and it pushes the bottom of the page down" is sort of jarring, but not a blocker.

src/util/related-docs.ts
```diff
@@ -138,3 +138,19 @@ export function relatedUserGuideFor(
 ): RelatedLink[] {
   return rankedCandidates(current, allDocs).slice(0, HEADLESS_MAX).map(toLink)
 }
+
+/**
+ * Other user-guide pages carrying a given tag, sorted alphabetically by title.
+ * Used by the clickable-tags UI to populate per-tag expansion content.
+ * Excludes the current page so a tag never lists itself.
+ */
+export function userGuidePagesWithTag(
```
zastrowm, line 150: Why alphabetically?  I would have thought "most relevant"/"highest scoring" first?

src/components/PageTags.astro
```diff
@@ -0,0 +1,265 @@
...
+
+const { starlightRoute } = Astro.locals
+const entry = starlightRoute.entry
+
+const allDocs = entry.collection === 'docs' ? await getCollection('docs') : []
+const tags: string[] = entry.collection === 'docs' ? (entry.data.tags ?? []) : []
+
+// Build a panel id that's unique to this page + tag, and safe for HTML
+// attribute use even if a tag ever contains characters outside [a-z0-9-].
+const idSlug = (s: string) => s.replace(/[^a-zA-Z0-9-]/g, '-')
```
zastrowm: I thought we might have a slugger utility somewhere already

zastrowm (dismissed):

mehtarac (dismissed):

src/components/overrides/MarkdownContent.astro
notowen333: I think the bottom makes the most sense. 

The sidebar doesn't have enough screen real estate and then we'd mix up an index with something completely different.

The top is distracting attention too early; this feature is supposed to give optional branching, not an initial touch point.

src/util/related-docs.ts
notowen333: Fair — switched to relevance score in 8bbed47. Reusing the same specificity-weighted Jaccard scorer used for the headless Related Pages list, so pages sharing more topical context with the current page rank above ones that only carry the queried tag. Ties still break alphabetically for determinism.

src/components/PageTags.astro
notowen333: Yep — `github-slugger` is already imported in `src/content.config.ts`. Switched to it in 8bbed47.

src/components/PageTags.astro
notowen333: Tried this and decided it doesn't work with the allotted space. Prefer the before approach where we add vertical space

mehtarac (approved):

PR #843, opened 2026-05-15
fix: eliminate open redirects in 404 redirect logic

pgrayy:
## Description

Fixes an open redirects in the 404 client-side redirect logic. The `/latest/` version prefix stripping could produce an absolute URL (e.g. `/latest/https://evil.com` becomes `https://evil.com`) which was then passed directly to `location.replace()`, redirecting users off-site.

Try it: https://strandsagents.com/latest/https://google.com

The fix adds an early return in `resolveRedirectFromUrl` that rejects paths starting with `https://` or `http://` after normalization. Explicit external redirects (discord, GitHub examples) still work because SLUG_RULES match on the slug (e.g. `discord`) before the path could ever look like a URL.

## Type of Change

- Bug fix

## Checklist

- [x] Try https://d3ehv1nix5p99z.cloudfront.net/pr-cms-843/https://google.com
- [x] Try https://d3ehv1nix5p99z.cloudfront.net/pr-cms-843/discord

----


astro.config.mjs
```diff
@@ -25,6 +25,15 @@ const sidebar = loadSidebarFromConfig(
 export default defineConfig({
   site: 'https://strandsagents.com',
   base: process.env.ASTRO_BASE_PATH || '/',
+  redirects: {
+    '/discord': 'https://discord.gg/strands',
```
zastrowm: FWIW, the reason we don't use these is because it ends up with a bright flash of white before redirecting.  It sort of sucks as a UX.  It *was* intentional FWIW

zastrowm (dismissed):

astro.config.mjs
pgrayy: Ack. See revision.

mehtarac (approved):

zastrowm (approved):

PR #844, opened 2026-05-15
design: add long term memory design

opieter-aws:
## Type of Change

- New content
- Content update/revision
- Typo/formatting fix
- Bug fix
- Other (please describe):

## Checklist

----


designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+    { store: userStore, namespace: 'user-123' },
+    { store: teamStore, namespace: 'team-marketing' },
+    { store: orgStore,  namespace: 'org-acme' },
+  ],
+})
+```
+
+Multi-store support avoids pushing multi-tenancy complexity onto the developer. A single agent can query personal, team, and organization knowledge simultaneously, with namespace isolation keeping them separate.
+
+**Read-only vs. mutable stores.** Two interfaces: `KnowledgeStore` (search-only) and `MutableKnowledgeStore` (search + write + delete). This distinction makes multi-tenant patterns natural: team or org stores that are pre-populated externally are read-only, while a user's personal store is mutable and accepts new facts during conversation. Mutability is determined by whether the store has an ingestion configuration (see Knowledge Ingestion below).
```
mkmeral: is `MutableKnowledgeStore` essentially CRUD interface then?

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+    tools: true, // Agent gets search_memory tool.
+  }),
+})
+```
+
+When multiple stores are configured, results are interleaved by rank using store config order as the priority signal. Scores are not compared across stores because different backends produce incomparable scales. If a store fails, partial results from other stores are still returned.
+
+#### Context Injection
+
+Context injection guarantees that relevant knowledge is always present, at the cost of paying for retrieval every turn. This is useful when baseline context is more important than token efficiency, or when the model can't reliably judge when to search.
```
zastrowm: Why did we decide on a system prompt vs a distinct message?

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+Active recall lets the agent decide when memory is relevant. Instead of retrieving knowledge every turn, the agent searches on demand, only when it judges that stored knowledge would help.
+
+This works by registering a `search_memory` tool that the agent can call like any other tool. The trade-off: active recall depends on the model recognizing when to search. If the model doesn't think to look, relevant memories stay hidden.
+
+```typescript
+const agent = new Agent({
+  model,
+  memoryManager: new MemoryManager({
+    stores: [{ store, namespace: 'user-123' }],
+    tools: true, // Agent gets search_memory tool.
```
mkmeral: is the tool opt-in? should it be? and why? 

I generally think agentic memory/RAG concepts perform better than raw input related heuristic memory approaches

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+
+### Minimal: prototyping
+
+```typescript
+import { Agent, MemoryManager, InMemoryKnowledgeStore } from '@strands-agents/sdk'
+
+const agent = new Agent({
+  model,
+  memoryManager: new MemoryManager({
+    stores: [{ store: new InMemoryKnowledgeStore(), namespace: 'user-123', ingestion: { trigger: 'tool' } }],
```
zastrowm: Where does `namespace` become relevent? Is it exposed to the user or for some other reason?

designs/0011-knowledge-bases.md
opieter-aws: The agent takes system prompt injection with higher priority than a message. We can run additional benchmarking here, but that's what research shows

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+**Read-only vs. mutable stores.** Two interfaces: `KnowledgeStore` (search-only) and `MutableKnowledgeStore` (search + write + delete). This distinction makes multi-tenant patterns natural: team or org stores that are pre-populated externally are read-only, while a user's personal store is mutable and accepts new facts during conversation. Mutability is determined by whether the store has an ingestion configuration (see Knowledge Ingestion below).
+
+**Shipped backends.**
+
+| Backend | Package | Use case |
+|---------|---------|----------|
+| `InMemoryKnowledgeStore` | `@strands-agents/sdk` | Testing, prototyping |
+| `FileKnowledgeStore` | `@strands-agents/sdk` | Local development |
+| `BedrockKnowledgeBaseStore` | `@strands-agents/sdk` | Production (managed, zero-infra) |
+| `AgentCoreKnowledgeStore` | `@strands-agents/memory-agentcore` | AgentCore managed memory |
```
notowen333: So this would be a new package that Strands owns instead of the existing AC SDK?

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+| `scheduled` | Every N turns (configurable via `interval`) | Controllable |
+
+All writes are async and non-blocking. This means a fact stored in one turn may not be searchable immediately in the next (eventual consistency).
+
+**Deduplication.** MemoryManager tracks a per-store high-water mark: a pointer to the last message that was already processed. Each trigger only processes messages beyond that mark, preventing duplicate writes. Tool-related content blocks (`toolUse`, `toolResult`) are filtered out by default before processing, since they rarely contain user-relevant knowledge.
+
+**Custom triggers.** For cases the built-in triggers don't cover, the store interface is public and can be called directly from any lifecycle hook.
+
+**Deletion and corrections.** `MutableKnowledgeStore.delete()` is exposed for programmatic use (compliance, cleanup), but no `delete_memory` tool is registered for the agent. Exposing deletion to the agent risks accidental data loss with no undo path. Instead, corrections are handled by storing updated facts. Newer entries take precedence via recency weighting in search results.
```
gautamsirdeshmukh: Is the idea that the knowledge store will be capped at N memories and "stale" memories will gradually be phased out over time (as new/updated memories come in)? Or, are you (also) envisioning some mechanism of correlating / squashing similar memories?

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+})
+```
+
+When multiple stores are configured, results are interleaved by rank using store config order as the priority signal. Scores are not compared across stores because different backends produce incomparable scales. If a store fails, partial results from other stores are still returned.
+
+#### Context Injection
+
+Context injection guarantees that relevant knowledge is always present, at the cost of paying for retrieval every turn. This is useful when baseline context is more important than token efficiency, or when the model can't reliably judge when to search.
+
+Enabled via `injection: true`. Each turn, MemoryManager searches stores using the last substantive user message as the query, formats results into a `<strands-memory>` block, and appends it to the system prompt. The block is stripped and re-injected fresh every turn so memories never accumulate in the prompt and each turn gets a fresh set based on the current query. The token budget and formatting are configurable via `maxTokens`, `format`, and `query` functions.
```
mkmeral: > using the last substantive user message as the query

what does this mean? what is the last "substantive" user message?

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+    injection: true, // searches every turn, injects into system prompt
+  }),
+})
+```
+
+---
+
+### Knowledge Ingestion
+
+Knowledge can enter the system in two ways: the agent explicitly writes it (via a `store_memory` tool registered by MemoryManager), or MemoryManager automatically extracts it from conversation messages using an extractor (a component that distills messages into discrete facts via a model call).
```
mkmeral: should store/search memory be separate tools? can't we do dynamic tooling instead? just have a `memory` tool with `search` and `store` as actions? 🤔

designs/0011-knowledge-bases.md
notowen333: nit: would workshop the config UX more. `tools: true` is surprising to me

designs/0011-knowledge-bases.md
mkmeral: as we split the tools more, it'll cause pain/annoyance (on toolspec and token side + tool shortlisting side [lke manage tools])

that said if it performs better on evals, sure

designs/0011-knowledge-bases.md
notowen333: Maybe like retrieveTools and writeTools could be more granular/clear

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+  memoryManager: new MemoryManager({
+    stores: [{ store, namespace: 'user-123', ingestion: { trigger: 'tool' } }],
+    injection: true,  // default XML format, 2000 token budget
+  }),
+})
+```
+
+## Alternatives Considered
+
+### 1. Memory as a top-level Agent parameter (no MemoryManager class)
```
zastrowm: We should be viewing config as a streamlined version of the classes.  For example, this looks much more concise than creating a class and I tend to like this more.

Typing would be `memory: TheConfigType | TheClass`

Of course we should consider the 80/20 case for deciding what we expose but we don't have to decide one or the other

designs/0011-knowledge-bases.md
opieter-aws: Currently leaning towards defaulting tools enabled, but defaults I want to only confirm after benchmarking.

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+
+Knowledge can enter the system in two ways: the agent explicitly writes it (via a `store_memory` tool registered by MemoryManager), or MemoryManager automatically extracts it from conversation messages using an extractor (a component that distills messages into discrete facts via a model call).
+
+MemoryManager uses triggers to control when these writes happen. A trigger is a named event that causes MemoryManager to process recent messages and write to the store. Four built-in triggers cover most use cases:
+
+| Trigger | When | Cost |
+|---------|---------------|---------------|
+| `tool` | Agent calls `store_memory` | Nothing extra (agent provides content directly) |
+| `perTurn` | After every agent invocation | High (model call per turn if an extractor is configured) |
+| `onEviction` | Messages are evicted from the context window | Medium (only fires on eviction events) |
```
mkmeral: how do we keep track of this? how do i know when messages are evicted?

designs/0011-knowledge-bases.md
poshinchen: Could we control them via args? (e.g. mutable: bool, or rules: list[str]).

opieter-aws: Correct

designs/0011-knowledge-bases.md
lizradway: I think this is leaving any optional eventual cleanup or compliance strategy to the dev, so deletion is available as a necessity, but not shipping it as a part of the built in mechanism accessible to the agent.

I'm a big fan of this approach + it seems to be pretty standard.

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+const agent = new Agent({
+  model,
+  memoryManager: new MemoryManager({
+    stores: [{ store, namespace: 'user-123', ingestion: { trigger: 'tool' } }],
+    injection: true,  // default XML format, 2000 token budget
+  }),
+})
+```
+
+## Alternatives Considered
```
zastrowm: Did we consider thinking about memories as files instead of in-memory representations?

designs/0011-knowledge-bases.md
opieter-aws: `namespace` is used by the store to determine the search scope. Each user will have their own namespace within the same store for example.

designs/0011-knowledge-bases.md
opieter-aws: No this will be owned by AC, disregard the agentic naming

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+
+Extraction is optional. It is only needed when two conditions are true: the backend doesn't handle extraction server-side, and you want MemoryManager to distill conversations into facts automatically rather than relying on the agent to provide facts via the `tool` trigger.
+
+Each store can have its own extractor, because different stores benefit from different extraction styles. A preferences store might want pure facts ("user prefers dark mode") while a decisions store wants richer context with reasoning. `ModelExtractor` is the built-in implementation: it calls a language model to extract facts, defaulting to the agent's own model but configurable with an explicit cheaper model to reduce cost.
+
+When no extractor is configured, messages are serialized as plain text and passed directly to the store's `add()` method. This is the correct setup for managed backends that handle extraction internally.
```
mkmeral: so this is when a store is mutable but doesn't have extractor?

designs/0011-knowledge-bases.md
lizradway: I am adding a new `onEviction` event to conversation/context managers :)

designs/0011-knowledge-bases.md
opieter-aws: We will not own the knowledge store - we only provide search and add interfaces. How the knowledge store is managed, is determined by the pluggable backend

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+### What Becomes Easier
+
+- Cross-session knowledge becomes a single parameter. No custom persistence, no manual tool registration, no vector store wiring.
+- Multi-tenancy is built in via multi-store + namespacing.
+- Progressive complexity: `InMemoryKnowledgeStore` (prototyping) to `BedrockKnowledgeBaseStore` (production) is changing one import.
+
+### What Becomes Harder or Requires Attention
+
+- **Eventual consistency**: writes are async; a fact may not be searchable in the next turn.
+- **Extraction cost**: `perTurn` triggers a model call every turn. We need sensible defaults and good documentation for users to navigate this.
```
zastrowm: How does this compare generally to what other frameworks use?  And/or how do most services handle this?  Are they mostly doing this server side anyways?

designs/0011-knowledge-bases.md
poshinchen: +, do we want to allow users to provide their tools?

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+
+Extraction is optional. It is only needed when two conditions are true: the backend doesn't handle extraction server-side, and you want MemoryManager to distill conversations into facts automatically rather than relying on the agent to provide facts via the `tool` trigger.
+
+Each store can have its own extractor, because different stores benefit from different extraction styles. A preferences store might want pure facts ("user prefers dark mode") while a decisions store wants richer context with reasoning. `ModelExtractor` is the built-in implementation: it calls a language model to extract facts, defaulting to the agent's own model but configurable with an explicit cheaper model to reduce cost.
+
+When no extractor is configured, messages are serialized as plain text and passed directly to the store's `add()` method. This is the correct setup for managed backends that handle extraction internally.
+
+## Developer Experience
```
mkmeral: what are the defaults? you enabled both direct ingestion and tools with a param, so we do we need customers to enable one or the other?

designs/0011-knowledge-bases.md
opieter-aws: See https://github.com/strands-agents/docs/pull/844/changes#diff-bbb536e8ceda86b8e7bbc8a665ee8f4eca2ac0d5ead2259f6c043ded15e54d1fR366

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+</details>
+
+---
+
+<details>
+<summary><b>Appendix B: Implementation Details</b></summary>
+
+### Context injection lifecycle
+
+When injection is enabled, MemoryManager hooks into `BeforeInvocationEvent` (once per turn). The lifecycle:
```
zastrowm: Side-topic, but we should push for a better way to do this IMHO; the whole "modify the system-prompt on every term" seems like a non-ideal hack at this point

designs/0011-knowledge-bases.md
opieter-aws: If they can be registered separately, we can consider that. We shouldn't couple them in case of a readOnly store.

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+
+### What Becomes Harder or Requires Attention
+
+- **Eventual consistency**: writes are async; a fact may not be searchable in the next turn.
+- **Extraction cost**: `perTurn` triggers a model call every turn. We need sensible defaults and good documentation for users to navigate this.
+- **Active recall and context injection depends on model judgment**: the model must know when to search. Context injection guarantees baseline context at the cost of always paying for retrieval. We need to evaluate and baseline tool descriptions.
+
+### Migration
+
+No breaking changes. `memoryManager` is a new optional parameter on `AgentConfig`.
```
mkmeral: how does this relate to conversation manager, session manager, snapshots?

designs/0011-knowledge-bases.md
awsarron: consider multimodal

designs/0011-knowledge-bases.md
opieter-aws: See https://github.com/strands-agents/docs/pull/844/changes#diff-bbb536e8ceda86b8e7bbc8a665ee8f4eca2ac0d5ead2259f6c043ded15e54d1fR305 . AFAIK we're using the same setup for context offloading. It's a situation of `boolean` | `defaults`

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+
+Context injection guarantees that relevant knowledge is always present, at the cost of paying for retrieval every turn. This is useful when baseline context is more important than token efficiency, or when the model can't reliably judge when to search.
+
+Enabled via `injection: true`. Each turn, MemoryManager searches stores using the last substantive user message as the query, formats results into a `<strands-memory>` block, and appends it to the system prompt. The block is stripped and re-injected fresh every turn so memories never accumulate in the prompt and each turn gets a fresh set based on the current query. The token budget and formatting are configurable via `maxTokens`, `format`, and `query` functions.
+
+```typescript
+const agent = new Agent({
+  model,
+  memoryManager: new MemoryManager({
+    stores: [{ store, namespace: 'user-123' }],
```
notowen333: Where do we imagine defining permissioning for namespaces?

designs/0011-knowledge-bases.md
opieter-aws: Correct, it's the case for e.g. BedRock KBs and in memory / file system setup.

designs/0011-knowledge-bases.md
zastrowm: Why is this not on the store in that case?

designs/0011-knowledge-bases.md
opieter-aws: We will be tracking a high water mark in the memory manager. Message eviction by the context manager will trigger the `onEviction` hook.

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+Under the hood, MemoryManager integrates with the agent lifecycle via hooks: registering tools at initialization, injecting knowledge before model calls, and ingesting new facts after each turn.
+
+**Stores.** A store is a backend that holds and retrieves knowledge (a vector database, a managed service like Amazon Bedrock Knowledge Bases or AgentCore Memory, or any implementation of the store interface). MemoryManager orchestrates one or more stores, each scoped by a namespace:
+
+```typescript
+memoryManager: new MemoryManager({
+  stores: [
+    { store: userStore, namespace: 'user-123' },
```
poshinchen: Should we type the `Store` class?

designs/0011-knowledge-bases.md
mkmeral: lol I have the same mental model of keeping system prompt for instructions only, but the industry tends to dump the data to system prompt. That keeps conversational flow normal (sequential), and it makes system prompt into this dynamic component.

mkmeral: qq, given that we keep appending xml to system prompt (skills, memory, etc),  should we make it more configurable? should we improve our system prompt handling? 🤔

designs/0011-knowledge-bases.md
zastrowm: From our offline description; I do think we should consider renaming to "id" or "name" or "storeName"; something that indicates it's the lookup to the name

designs/0011-knowledge-bases.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+### Production: Bedrock Knowledge Bases
+
+```typescript
+import { Agent, MemoryManager, BedrockKnowledgeBaseStore, ModelExtractor } from '@strands-agents/sdk'
+
+const agent = new Agent({
+  model,
+  memoryManager: new MemoryManager({
```
notowen333: It feels to me that we might be well served to attempt implementing a couple of concrete 3P backends while defining our extractor/store/manager re-usable pieces.

Just to make sure we are aligned with the concrete use cases

designs/0011-knowledge-bases.md
notowen333: Can we break up the action types and then one construct has a property which lists which types:

i.e. only delete

Instead of two diverging interfaces what have pre-defined actions

designs/0011-knowledge-bases.md
opieter-aws: We're keeping multimodal out of scope for the first implementation, and support it as follow-up

designs/0011-knowledge-bases.md
opieter-aws: Decided to make multi-tenancy a responsibility of the store. Stores are scoped to their namespace/scope upon instantiation.

designs/0011-knowledge-bases.md
opieter-aws: Decided to make multi-tenancy a responsibility of the store. Stores are scoped to their namespace/scope upon instantiation, and it can use the term which-ever is used by the back-end.

designs/0011-knowledge-bases.md
opieter-aws: Thanks for the suggestion, did the POC implementation for AgentCore and HindSight and it surfaced some good findings!

designs/0011-knowledge-bases.md
opieter-aws: Updated in latest version

designs/0011-knowledge-bases.md
opieter-aws: Decided to go with this approach in the updated design

designs/0011-knowledge-bases.md
opieter-aws: It's typed, in the updated design there's just 1 interface `KnowledgeStore` which has optional methods

notowen333 (approved):

PR #848, opened 2026-05-19
feat(tabs): hide inline language tablist; header toggle is sole selector

agent-of-mkmeral:
## Summary

Hide the **inline** Python/TypeScript tab selector that Starlight renders at every language tab group. The header `LanguageToggle` becomes the single, site-wide language selector.

**All tab logic is preserved** — only the inline UI is hidden. The header toggle continues to drive panel switching exactly as before.

Fixes the distraction reported in #839 (and similar pages with many tab groups), where every section gets its own visible Python/TypeScript bar.

## Why

We currently have two visible selectors for language:

1. **Header `LanguageToggle`** — site-wide, sticky
2. **Inline tablist** — rendered by Starlight `<Tabs>` at every tab group

On dense pages (e.g. PR https://github.com/strands-agents/docs/pull/839, `hooks.mdx`, `custom-tools.mdx`), the inline bars repeat dozens of times and become visual noise. Both selectors are already kept in sync via Starlight's synced-tabs mechanism (`starlight-synced-tabs__jarkqt`), so the inline UI is redundant.

## Approach

All changes are in **one file**: `src/components/AutoSyncTabs.astro`.

1. Detect language tabs by inspecting the rendered slot for labels exactly equal to `{"Python", "TypeScript"}` (in any order).
2. When matched, wrap the rendered `<Tabs>` in a `<div data-language-tabs>` marker.
3. A scoped global CSS rule hides `starlight-tabs > .tablist-wrapper` inside that marker.
4. Non-language tab groups (macOS/Windows, npm/yarn, IDE choice, AWS App Runner / Lambda / Cloud Run, etc.) are **untouched** — they still need their inline tablist.

## Why the logic still works

Looking at `LanguageToggle.astro`:

```js
function syncStarlightTabs(lang) {
  document.querySelectorAll('starlight-tabs').forEach(function (tabGroup) {
    var tabs = tabGroup.querySelectorAll('[role="tab"]');
    tabs.forEach(function (tab) {
      if (tab.textContent.trim() === lang && tab.getAttribute('aria-selected') !== 'true') {
        tab.click();
      }
    });
  });
}
```

The header toggle calls `tab.click()` on `[role="tab"]` elements inside `<starlight-tabs>`. Starlight's `switchTab()` then:

- toggles `aria-selected` on tabs
- sets `hidden` on tabpanels
- syncs all other `<starlight-tabs>` with the same `data-sync-key`

None of that depends on the tablist being **visible** — `display: none` doesn't disable click handlers, doesn't break the DOM tree, and doesn't affect the synced-tabs registry. The `starlight-tabs-restore` script also still runs because it operates on the DOM directly, not visibility.

## What stays the same

- ✅ `AutoSyncTabs` syncKey hash (`jarkqt` for `Python|TypeScript`)
- ✅ Header `LanguageToggle` switching logic
- ✅ localStorage persistence (`starlight-synced-tabs__jarkqt`)
- ✅ Cross-tab/window storage events
- ✅ Paired-page navigation (quickstart, API reference)
- ✅ All non-language tab groups (npm/yarn, OS variants, IDE, etc.)
- ✅ Zero content migration — every existing `<Tabs>...<Tab label="Python">...<Tab label="TypeScript">` page just loses its tab bars

## What changes

- ❌ Inline tablist no longer **renders visually** for Python/TypeScript pairs
- The layout collapses cleanly because Starlight's CSS already gates the `margin-top: 1rem` on tabpanels via `.tablist-wrapper ~ :global([role='tabpanel'])` — when the wrapper has `display: none`, the sibling selector still applies (display does not remove the element), so spacing is preserved.

## Testing

- The change is purely CSS-driven (one DOM wrapper + one rule). No new tests required.
- Existing tests in `test/` don't reference `AutoSyncTabs` or `LanguageToggle`, so nothing to update.
- Visual verification: any page with `<Tabs><Tab label="Python">...<Tab label="TypeScript">` should show no inline tab bar; the header toggle should still switch them.
- Non-language tabs (e.g. `Linux/macOS/Windows` in deploy guides) should still display their tablist normally.

## Files changed

- `src/components/AutoSyncTabs.astro` (+44 / −3)

## Related

- Issue: agent-of-mkmeral/strands-coder-private#208
- Trigger: https://github.com/strands-agents/docs/pull/839 (example of distracting density)
- Builds on the language-toggle work that introduced the header selector.

### DIFF
--- a/src/components/AutoSyncTabs.astro
+++ b/src/components/AutoSyncTabs.astro
@@ -10,6 +10,13 @@
  *   <TabItem label="npm">npm install</TabItem>
  *   <TabItem label="yarn">yarn add</TabItem>
  * </AutoSyncTabs>
+ *
+ * Special case: language tabs.
+ * When the labels are exactly the language pair {"Python", "TypeScript"},
+ * the wrapper is marked with `data-language-tabs` so the inline tablist
+ * can be visually hidden via global CSS. The site's header LanguageToggle
+ * remains the single language selector. Tab logic, syncing, and switching
+ * are unchanged — only the inline UI is hidden.
  */
 import { Tabs } from '@astrojs/starlight/components';
 
@@ -43,8 +50,42 @@ function hashString(str: string): string {
 
 // Generate syncKey from labels if not explicitly provided
 const syncKey = explicitSyncKey ?? (labels.length > 0 ? hashString(labels.join('|')) : undefined);
+
+// Detect language tabs (Python + TypeScript pair, in any order).
+// Only this exact pair is treated as a "language tab" — other tab groups
+// (npm/yarn, OS variants, IDE choice, etc.) keep their normal tablist UI.
+const labelSet = new Set(labels);
+const isLanguageTabs =
+  labels.length === 2 &&
+  labelSet.has('Python') &&
+  labelSet.has('TypeScript');
 ---
 
-<Tabs syncKey={syncKey}>
-  <Fragment set:html={slotContent} />
-</Tabs>
+{isLanguageTabs ? (
+  <div data-language-tabs>
+    <Tabs syncKey={syncKey}>
+      <Fragment set:html={slotContent} />
+    </Tabs>
+  </div>
+) : (
+  <Tabs syncKey={syncKey}>
+    <Fragment set:html={slotContent} />
+  </Tabs>
+)}
+
+<style is:global>
+  /*
+   * Hide the inline tablist for language tabs only.
+   * The header LanguageToggle is the single source of truth for language
+   * selection — inline tablists became visually distracting on pages with
+   * many language tab groups. The Starlight tab logic is preserved (panels
+   * are still selectively shown/hidden by Starlight), so the header toggle
+   * can programmatically switch tabs via tab.click() exactly as before.
+   *
+   * We keep the inline tablist for non-language tabs (e.g. macOS/Windows,
+   * npm/yarn) — those still need an inline selector.
+   */
+  [data-language-tabs] starlight-tabs > .tablist-wrapper {
+    display: none;
+  }
+</style>
--- a/src/components/overrides/Header.astro
+++ b/src/components/overrides/Header.astro
@@ -59,6 +59,15 @@ const { siteTitleHref } = Astro.locals.starlightRoute;
             <Icon name="down-caret" class="mobile-nav-caret" size="1rem" />
           </summary>
           <div class="mobile-nav-menu">
+            {/* Language toggle at the top of the dropdown.
+                 Most-used control for the docs (Python vs TypeScript) so it
+                 takes the prime spot. The header bar stays uncluttered on
+                 narrow viewports because the toggle no longer lives there. */}
+            <div class="mobile-nav-section-label mobile-nav-section-label--first">Language</div>
+            <div class="mobile-nav-language">
+              <LanguageToggle />
+            </div>
+            <div class="mobile-nav-divider"></div>
             {navLinks.map((link: NavLink) => {
               const isActive = activeNav?.label === link.label;
               return (
@@ -72,14 +81,6 @@ const { siteTitleHref } = Astro.locals.starlightRoute;
                 </a>
               );
             })}
-            {/* Language toggle inside the mobile dropdown menu.
-                 Keeps the header bar uncluttered on narrow viewports while
-                 still giving mobile users access to switch languages. */}
-            <div class="mobile-nav-divider"></div>
-            <div class="mobile-nav-section-label">Language</div>
-            <div class="mobile-nav-language">
-              <LanguageToggle />
-            </div>
             {githubSections.map((section: GitHubSection, i: number) => (
               <Fragment>
                 <div class={i === 0 ? 'mobile-nav-divider' : undefined}></div>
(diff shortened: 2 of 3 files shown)


src/components/AutoSyncTabs.astro
```diff
@@ -43,8 +50,42 @@ function hashString(str: string): string {
...
+   * The header LanguageToggle is the single source of truth for language
+   * selection — inline tablists became visually distracting on pages with
+   * many language tab groups. The Starlight tab logic is preserved (panels
+   * are still selectively shown/hidden by Starlight), so the header toggle
+   * can programmatically switch tabs via tab.click() exactly as before.
+   *
+   * We keep the inline tablist for non-language tabs (e.g. macOS/Windows,
+   * npm/yarn) — those still need an inline selector.
+   */
+  [data-language-tabs] starlight-tabs > .tablist-wrapper {
```
zastrowm, line 88: The global language switcher isn't on mobile right now which is why we didn't do this initially.

(view https://d3ehv1nix5p99z.cloudfront.net/pr-cms-848/docs/user-guide/concepts/tools/ on mobile).

The API References also have this issue, but I don't want to make it worse

mkmeral: @agent-of-mkmeral  let's update this. bring the header selector to mobile (consider how it would actually render)

agent-of-mkmeral: Done — pushed [180df8f](https://github.com/strands-agents/docs/pull/848/commits/180df8f1).

**Summary**: moved `<LanguageToggle />` out of the desktop-only `.right-group` (which is hidden on mobile via `sl-hidden md:sl-flex`) into a new always-visible `.header-language-toggle` wrapper, placed after the search bar and before the mobile-nav menu.

**Mobile rendering** (thinking through how it actually fits):

Mobile header is tight — already has logo (180px) + search + hamburger. To make the toggle fit, on `<50rem` viewports it collapses to **icon-only**:
- Python and TypeScript SVG icons stay visible
- Text labels hidden via the standard clip-based visually-hidden technique (still in the DOM, still announced by screen readers)
- `aria-label="Python"` / `aria-label="TypeScript"` added on each radio button so AT users still get the language name
- Button padding tightened to `0.25rem 0.4rem` and the icon↔label gap removed for a compact pill

**Mobile layout becomes:** `[logo] [search] [🐍][TS] [☰]`

**Desktop unchanged**: text labels still show, full toggle still in the right-side header controls group.

The `50rem` breakpoint matches Starlight's existing mobile/desktop boundary (same one used by `sl-hidden` / `md:sl-flex` elsewhere in the header), so nothing else shifts.

Verified with `astro dev`: rendered markup contains the new wrapper without `sl-hidden`, the mobile media query is emitted in CSS, and tabs pages still get `data-language-tabs` so the inline tablist is hidden as before. Preview should redeploy shortly — please re-check on mobile.

@zastrowm on the API References concern: this PR's CSS only hides the inline tablist for `[data-language-tabs]` (i.e. `<AutoSyncTabs>` with the `{Python, TypeScript}` pair). API Reference pages don't go through `AutoSyncTabs`, so they're unaffected here. Once the header toggle is reachable on mobile (this commit), we can revisit those pages separately if you'd like.

mkmeral (approved):

zastrowm (approved):

src/components/overrides/Header.astro
```diff
@@ -465,22 +481,45 @@ const { siteTitleHref } = Astro.locals.starlightRoute;
...
+     * On desktop, switch header-top from flexbox to CSS Grid for precise control.
+     * mobile-nav is display:none on desktop and removed from grid flow, so the
+     * visible tracks are:
+     *
+     *   Column 1 (180px): Logo - fixed width prevents layout shift
+     *   Column 2 (1fr):   Spacer - flexible, pushes remaining items to the right
+     *   Column 3 (300px): Search - fixed width for consistent positioning
+     *   Column 4 (auto):  Right group - social icons, theme, language toggle
      */
     @media (min-width: 50rem) {
```
zastrowm, line 515: As a follow-up (not this pr!) I wonder if we could iterate on where the switcher goes. It's still a little clunky but I wonder if it could exist in the outline (2nd header line) in a condensed form:

zastrowm (approved):

src/components/overrides/Header.astro
mkmeral: @agent-of-mkmeral  create another PR for this. also check if there are any libraries, aws services, etc. that you might have access to, so you can render the website and see it in github action runtime. so you can check these things visually before returning

agent-of-mkmeral: Got it on both fronts.

**1. New PR for @zastrowm's outline-condensed idea** — spinning up now in a separate branch. Will explore moving the switcher into the second header row (outline area) per the screenshot, keeping this PR scoped to its current behavior.

**2. Visual verification in the GitHub Actions runtime** — yes, this is achievable. Quick scan of what's already on the `ubuntu-latest` runner I'm executing inside:

| Tool | Status | Use |
|------|--------|-----|
| `google-chrome` / `chromium` / `firefox` | ✅ pre-installed at `/usr/bin/` | headless render |
| Node 20 + `npx` | ✅ | `astro build` + `astro preview` |
| Playwright | pip-installable (~30s) | scripted screenshots at multiple viewports |
| AWS Bedrock (Claude multimodal) | ✅ creds present (`sts get-caller-identity` succeeds) | I can pipe the PNG back into my own context to actually *see* the rendered page |
| ImageMagick `convert` | apt-installable | side-by-side / diff |
| ~17 GB free disk on runner | sufficient for an `astro build` |

**Verification loop I'll wire into the new PR:**
1. `npm ci && npm run build && npm run preview &` to spin up a local server inside the runner.
2. Playwright script: load a tabs-heavy page (e.g. `/docs/user-guide/concepts/tools/`) at three viewports — `375×812` (mobile), `768×1024` (tablet), `1280×800` (desktop). Screenshot each.
3. Pipe each PNG through Bedrock multimodal so I can visually inspect the output before I commit — same loop you'd run locally with browser devtools, just remote.
4. Iterate on the outline-condensed layout until it looks right at all three breakpoints, *then* push.

I'll factor the screenshot helper into a small reusable script (`scripts/screenshot-preview.ts` or similar) so future UI PRs can use the same visual-verification gate. Will reference the new PR back in this thread when it's up.

PR #850, opened 2026-05-20
Design: Managing forked WASM dependencies

pgrayy:
Proposes using git subtrees with slim branches to house forked WASM dependencies (wasmtime, wasmtime-py, componentize-js, jco) within sdk-typescript. Includes size analysis and comparison with alternatives (submodules, LFS).

### DIFF
--- a/designs/0011-fork-management.md
+++ b/designs/0011-fork-management.md
@@ -1,0 +1,174 @@
+# Managing Forked Dependencies
+
+## Overview
+
+The WASM integration in sdk-typescript depends on modified versions of several upstream repositories:
+
+- **wasmtime** (Rust) — the WASM runtime
+- **wasmtime-py** (Python) — Python bindings for wasmtime
+- **componentize-js** (JavaScript) — JS component model tooling
+- **jco** (JavaScript/Rust) — JS toolchain for WebAssembly Components
+
+This document proposes an approach for housing these forks within the sdk-typescript repo to streamline development and CI/CD.
+
+## Problem
+
+Upstream WASM tooling does not yet support all the features we need for the Python-via-WASM bindings. Certain capabilities (e.g., async streaming) require workarounds that live in forked versions of these dependencies. Today, those forks exist as separate repositories, which creates friction:
+
+- Changes that span sdk-typescript and a fork require coordinating across multiple repos, branches, and PRs.
+- CI/CD pipelines must clone and build from multiple sources, adding complexity and fragility to the build.
+- Contributors need to discover and set up the correct fork versions manually.
+- There is no single place to see the full picture of what's modified and why.
+
+Bringing the forks into sdk-typescript simplifies the development loop: one clone, one set of commits, one CI pipeline.
+
+## Solution
+
+Use **git subtrees** to import each fork's source directly into sdk-typescript.
+
+Subtrees copy a remote repo's file tree into a subdirectory of the host repo. After import, the files are regular tracked content. No special tooling is required beyond the `git subtree` command, which ships with git.
+
+Benefits of this approach:
+
+- **Single clone** — contributors get everything with a standard `git clone`. No extra steps.
+- **Atomic commits** — a change to sdk-typescript and a fork workaround land in one commit/PR.
+- **Upstream sync** — `git subtree pull` merges new upstream changes; `git subtree push` extracts patches back when ready to upstream.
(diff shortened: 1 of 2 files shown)


designs/0011-fork-management.md
```diff
@@ -0,0 +1,174 @@
...
+- Changes that span sdk-typescript and a fork require coordinating across multiple repos, branches, and PRs.
+- CI/CD pipelines must clone and build from multiple sources, adding complexity and fragility to the build.
+- Contributors need to discover and set up the correct fork versions manually.
+- There is no single place to see the full picture of what's modified and why.
+
+Bringing the forks into sdk-typescript simplifies the development loop: one clone, one set of commits, one CI pipeline.
+
+## Solution
+
+Use **git subtrees** to import each fork's source directly into sdk-typescript.
```
zastrowm, line 27: The biggest con to me is that this source code ends up in the history forever; e.g. there's no way to strip it later without rewriting history.  Whereas a submodule being a pointer [effectively] means the overhead is trivial once we remove them

chaynabors: That seems like a feature to me though no? At one point we vended these dependencies and getting old builds working means simply checking out the old history. Hermetic builds.

pgrayy: That is a good call out. It would permanently alter the size of sdk-typescript. Based on the metrics, it would end up being roughly 30MB (calculated from the WASM repo .git sizes) of dead weight once removed.

To mitigate, we could try slimming down wasmtime and jco even more than what was proposed.

zastrowm: Yeah, what Patrick said; it's a permanent size bump for what is/should be a temporary solution

chaynabors: So no subtree? Commit your fork into the repo without history, and we can override from your fork if we need to downstream anything?

chaynabors: In my experience submodules are just incredibly clunky and make it difficult to use git tools to split history or something. Memory is fuzzy here. My understanding is that they'd poison our repo and be hard to remove but don't take my word for it.

zastrowm: > Commit your fork into the repo without history, and we can override from your fork if we need to downstream anything?

Yeah effectively

> In my experience submodules are just incredibly clunky

If you're doing active development and changing things often, this is true - it's why I don't suggest it as an alternative to a monorepository.  But for infrequent changes or "pinning" at a specific commit, it works pretty well.

AFAIK there's no poisoning as it's more like a file/pointer more than anything. Deleting them is a couple of configs, whereas subtrees you have the entire history as part of your repo forever.

chaynabors: Ignore me, conflating LFS and submodules for some reason. Need a nap.

designs/0012-wasm-dependency-distribution.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+The `dev:link-forks` script in the root `package.json`:
+
+```json
+{
+  "scripts": {
+    "dev:link-forks": "npm link ./forks/componentize-js ./forks/jco"
+  }
+}
```
pgrayy, line 142: Could continue to add to these commands over time to offer more conveniences if need be. But I feel something like this is a reasonable start.

designs/0012-wasm-dependency-distribution.md
```diff
@@ -0,0 +1,221 @@
...
+- Publishing fork artifacts reliably.
+- Testing fork changes against sdk-typescript before shipping.
+- Giving developers a fast local iteration loop when modifying forks.
+- Keeping sdk-typescript's repo and CI simple.
+
+## Solution
+
+Publish fork artifacts to **npm** and **PyPI** under the `@strands-agents` scope. Use a dedicated fork repo with CI that tests against sdk-typescript before publishing. Use **git submodules** in sdk-typescript as an optional local development convenience.
+
+### Distribution
```
zastrowm, line 27: The biggest pro is that this streamlines development for those who don't need changes to these packages.  The biggest con is that it is a much higher bar for those that are making the changes to the packages.

How confident are we that changes will be infrequent?

designs/0012-wasm-dependency-distribution.md
zastrowm: It balances the development flow (with friction) with the consumer flow quite well IMHO

